PR #6, opened 2016-01-04
Removed the gid and nid from all the Model Classes as it overrides th…

NourhanOmar:
…e values of the ModelBase class

Replaced elasticsearch method with the find method in the gridhealthchecker

### DIFF
--- a/lib/JumpScale/data/models/Models.py
+++ b/lib/JumpScale/data/models/Models.py
@@ -152,6 +152,8 @@ class ModelGroup(ModelBase):
 
 
 class ModelJob(EmbeddedDocument):
+    nid = IntField(required=True)
+    gid = IntField(required=True)
     data = StringField(default='')
     streams = ListField(StringField())
     level = IntField()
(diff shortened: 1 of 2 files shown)


lib/JumpScale/data/models/Models.py
```diff
@@ -152,8 +152,6 @@ class ModelGroup(ModelBase):
 
 
 class ModelJob(EmbeddedDocument):
-    nid = IntField(required=True)
-    gid = IntField(required=True)
```
rkhamis: This one cannot be removed because it shouldn't be using the gid/nid of the current node, but that of where the job is running.

lib/JumpScale/data/models/Models.py
```diff
@@ -152,8 +152,6 @@ class ModelGroup(ModelBase):
 
 
 class ModelJob(EmbeddedDocument):
-    nid = IntField(required=True)
```
rkhamis: Discuss with @muhamadazmy whether the save occurs on the node that creates the Job object or it's saved centrally. If the latter, must stay

muhamadazmy: I don't agree, this is required to identify which agent ran the job. without it we can't tell.

Note: also notice that the `job` model doesn't inherit our BaseModel (because it's an embedded model) so it doesn't inherit nid or gid from base mode, hence we need to redefine.

PR #17, opened 2016-01-25
Pass service Object to ActionsBase class during instanciation

zaibon:
currently all the methods from the actions classes of AYS take as first argument a Service object on which the actions.
Since the service object instanciate the actions class, we can directly pass the service object into the actions base class.
This bring less confusion and simplify the interface of the action we implement in the actions class. The service object will always be available, no need to pass it as first argument of each methods.


zaibon:
@despiegk 
@rkhamis 
any comments on that ?

rkhamis:
I prefer it this way

rkhamis:
Maybe we should rename "serviceObj"  while we're at it?

zaibon:
yep, so just ```service``` ??

rkhamis:
I think that's cleaner, yes

PR #27, opened 2016-02-03
Tempbr

abdulrahmantkhalifa:
agent and agent controller work


lib/JumpScale/tools/cuisine_/Cuisine2.py
```diff
@@ -847,6 +847,13 @@ def file_md5(self, location):
     # =============================================================================
 
     @actionrun(action=False,force=False)
+    def joinpaths(self, *args):
+        path = ""
+        for arg in args:
+            path += "/%s"%arg
```
rkhamis: A little bit pointless

PR #40, opened 2016-02-14
Fixing Telegram

NourhanOmar:


lib/JumpScale/tools/telegram/TelegramBot.py
```diff
@@ -49,11 +49,14 @@ def start(self):
         """
         # self.api.process_updates()
         h=InteractiveHandler()
+        h.setActionsPath(path="%s/telegrambot/actions" % j.dirs.varDir)
```
rkhamis: Setting it here will always overwrite any user changes.

PR #58, opened 2016-02-21
added $binDir to path tmprrly to fix mongo install

abdulrahmantkhalifa:
if you dont want to merge dont its tmp fix i can keep it local and u can delete the branch


despiegk:
seems ok

PR #59, opened 2016-02-21
Jumpscripts with aggregator

NourhanOmar:


apps/agentcontroller/jumpscripts/jumpscale/backup_osis.py
```diff
@@ -1,39 +1,39 @@
...
+#         if not os.path.exists(spacepath):
+#             j.system.fs.createDir(spacepath)
+#         for category in osiscl.listNamespaceCategories(namespace):
+#             categoryfile = os.path.join(spacepath, category)
+#             try:
+#                 osiscl.export(namespace, category, categoryfile)
+#             except:
+#                 print 'Error exporting category: %s on namespace:%s' % (category, namespace)
             
-if __name__ == '__main__':
```
muhamadazmy: Please don't comment out code, just remove the entire file.

apps/agentcontroller/jumpscripts/jumpscale/disk_monitoring.py
```diff
@@ -26,6 +26,12 @@ def action():
     dcl = j.data.models.system.Disk
     stats = statsd.StatsClient()
     pipe = stats.pipeline()
+    hostname =j.sal.nettools.getHostname()
+    aggregator = j.tools.aggregator.getClient(j.core.db,  hostname)
```
muhamadazmy: it doesn't make sense to use the j.core.db because this redis instance listens on a unix-socket (not tcp),

Instead all monitoring scripts `MUST` receive a 'redis' argument which is the address of the redis instance you should use in the format 'host:port'

apps/agentcontroller/jumpscripts/jumpscale/disk_monitoring.py
```diff
@@ -85,6 +91,8 @@ def action():
                 odisk.save()
 
         for key, value in results.items():
+            aggregator.measure(tags=tags, key="disks.%s" % key, value=value, measurement="")
+
             pipe.gauge("%s_%s_disk_%s_%s" % (j.application.whoAmI.gid, j.application.whoAmI.nid, path, key), value)
```
muhamadazmy: This line should be removed. We are building a new system that is not backward compatible

muhamadazmy:
Please remove all the files that you commented out all its code.

muhamadazmy:
Also please remove the 'period' tag from the ALL the monitoring jumpscripts, we won't depend on the auto-scheduling anymore.

despiegk:
does the auto-scheduling still work in agent8? if yes maybe we should consider it
can also become cron if more useful, but how do we decide what is run on which agent? are the roles still there?

muhamadazmy:
@despiegk Yes, the auto scheduling still work in agent8, and yes the roles also still there. If a jumpscript has the period defined, the agentcontroller will auto schedule it on the agents that has the defined roles, or all agents if roles is omitted.

despiegk:
Cool lets use that then for the monitoring

muhamadazmy:
@despiegk 
Does this mean we don't need the reality capture defined here 
https://github.com/Jumpscale/Home/issues/3
?
I believe if we gonna use the controller feature to auto schedule jumpscripts with roles we don't need that anymore. right ?

despiegk:
dont understand

we need work #3 done

it will be scheduled by g8os_core automatically



On Mon, Feb 22, 2016 at 8:39 AM, Muhamad Azmy <notifications@github.com>
wrote:

> @despiegk <https://github.com/despiegk>
> Does this mean we don't need the reality capture defined here
> Jumpscale/Home#3 <https://github.com/Jumpscale/Home/issues/3>
> ?
> I believe if we gonna use the controller feature to auto schedule
> jumpscripts with roles we don't need that anymore. right ?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Jumpscale/jumpscale_core8/pull/59#issuecomment-187055661>
> .
>



-- 
------------------------
Kristof De Spiegeleer
+971525609014
+201206927877
+32 475405474
skype: despiegk

Issue #63, opened 2016-02-23
Model objects with the same guids are allowed


muhamadazmy:
If I created 2 objects (of the same type) using the models under j.data.models and
then set the guids of the 2 objects to the same value. Saving the 2 objects actually
creates 2 objects in the database.

I guess the guids must be unique and `indexed`. Overriding an object with another one that has the same guid must be possible.

muhamadazmy:
Mongo engine StringField has `unique` and `primary_key` flags that should be set on the guid field.

rkhamis:
`unique`, yes. the autogenerated id is our primary key, though. For beaker reasons.

mie00:
This uniqueness issue can be solved by adding `unique` flag.

But we can pass that because uuid v4 should be unique even if created across different machines.
Also there is a MongoEngine filed that is called `UUIDField` that ensures that the input is a uuid, and it has an option to store it as a binary object.

muhamadazmy:
Guids must also be primary keya

muhamadazmy:
@mie00 and @rkhamis 
My point is, although random guids generation grantee uniqueness (and primary key) of the guids, this must be explicitly set on the guid filed for the next reasons:

1- It will make getting an object with a guid much much faster.
2- If you explicitly set 2 objects to have the same guid, you can basically override the old object (I have a use case).

so If I created an object x with guid = 'xxx-xx-xxx' and then I created another object and set the same guid on the new object (with intention to override the old one) I should be able to completely override it just by using the same guid.

rkhamis:
Fixed

muhamadazmy:
@mie00 Can you also please add flag to do upsert in the model.save() function

mie00:
@muhamadazmy MongoEngine.save method supports upserting by default but only according to the primary key. It is possible that we expose model.save arguments to the Document.save method beside adding a special kwarg for upserting based on guid.

muhamadazmy:
I think (according to mongo docs) upsert takes a 'query' the query doesn't
have to be a primary key. So you we can do something like

pseudo code:
update({'guid': o.guid'}, o, upsert=true)

On Wed, Feb 24, 2016 at 9:51 AM, Mohamed I. Elawadi <
notifications@github.com> wrote:

> @muhamadazmy <https://github.com/muhamadazmy> MongoEngine.save method
> supports upserting by default but only according to the primary key. It is
> possible that we expose model.save arguments to the Document.save method
[...]

PR #64, opened 2016-02-23
make guid unique and indexed

mie00:
This is a fix for #63. I am not sure if the extending can be achieved in an easier way than using the \__new__ method.

### DIFF
--- a/lib/JumpScale/data/models/Models.py
+++ b/lib/JumpScale/data/models/Models.py
@@ -11,11 +11,34 @@ DB = 'jumpscale_system'
 class ModelBase():
     DoesNotExist = DoesNotExist
 
-    guid = StringField(default=lambda: str(uuid.uuid4()))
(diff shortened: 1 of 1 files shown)


rkhamis:
Could be done with a globally defined meta dict.
And whichever class needs to add to it, can copy it and do so

mie00:
meta dict can now be extended or created from scratch, depending on the class implementer's opinion.

lib/JumpScale/data/models/Models.py
```diff
@@ -7,15 +7,31 @@
 
 DB = 'jumpscale_system'
 
+default_meta = {'allow_inheritance': True, "db_alias": DB, 'indexes': ['guid']}
+
+def extend(a, b):
+    print(a,b)
```
grimpy: probably this print should be removed?

mie00: @grimpy I made a PR

PR #65, opened 2016-02-23
WIP: Refactoring

NourhanOmar:


lib/JumpScale/clients/portal/PortalFactory.py
```diff
@@ -0,0 +1,93 @@
...
+        addr = hrd.get('param.addr')
+        port = hrd.getInt('param.port')
+        secret = hrd.getStr('param.secret')
+        return self.get(addr, port, secret)
+
+    def get(self, ip="localhost", port=82, secret=None):
+        return Resource(ip, port, secret, "/restmachine")
+
+
+class PortalFactory():
```
grimpy: this class is not part of the client and does not belong in core, should stay in portal repo

PR #67, opened 2016-02-23
Bp update

zaibon:
Improvement of how AYS deal with modification of services.

When a modification is detected in a service. we need to find all impacted service. 
it's all the service that consume the service that has been modified.

you can use ```ays do action --printlonly``` to visualize then impact of the command before actually doing it.


despiegk:
when can we use this?

rkhamis:
We already can

despiegk:
best to merge into trunk then?

zaibon:
It has been merge back into master 4 days ago already.

PR #68, opened 2016-02-24
upsert document with the same guid

mie00:
#63 
Until now upsert does two queries to mongo. This happened because I didn't find a method in mongo that upserts the document and returns the id of the new document so I can bind it to mongoengine's Document.

### DIFF
--- a/lib/JumpScale/data/models/Models.py
+++ b/lib/JumpScale/data/models/Models.py
@@ -7,7 +7,7 @@ import uuid
 
 DB = 'jumpscale_system'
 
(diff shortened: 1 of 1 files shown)


mie00:
@muhamadazmy I have tested it and it worked as expected but with the limitation that the id can be an ObjectId or a valid string representation of it.

muhamadazmy:
@mie00 Does this mean we actually can't use a real guids ? or other strings that can't be converted to ObjectIds ?

mie00:
We cannot use any string that cannot be converted to an ObjectId. But if we created an id according to mongo's [ObjectID](https://docs.mongodb.org/manual/reference/object-id/) specs, it should be global as well.

PR #75, opened 2016-02-29
only sandbox /opt

zaibon:
Update jsdocker to only mount /opt/ when starting docker on top of fuse


zaibon:
@maxux  can you review the jsdocker and Docker lib see if I didn't forget anything.
Now only /opt needs to be mounted in containers.

maxux:
good for me

maxux:
seems okay for me

PR #94, opened 2016-03-10
Cuisine repair

zaibon:


lib/JumpScale/tools/actions/Action.py
```diff
@@ -491,16 +487,18 @@ def execute(self):
             output = ""
             counter=0
             ok=False
+            tb_text = ''
+            err = 'tb_text'
```
despiegk: why did we do this?

lib/JumpScale/tools/cuisine_/CuisineFactory.py
```diff
@@ -6,6 +6,7 @@ class OurCuisineFactory:
     def __init__(self):
         self.__jslocation__ = "j.tools.cuisine"
         self._local=None
+        self._cuinses_instance = {}
```
despiegk: typo cuinses

PR #97, opened 2016-03-10
decorator now shows docs and prams ,bug fix buildr

abdulrahmantkhalifa:


lib/JumpScale/tools/cuisine_/CuisineProcessManager.py
```diff
@@ -217,9 +217,10 @@ def __init__(self,executor,cuisine):
 
     def list(self,prefix=""):
         try:
+            import ipdb;ipdb.set_trace()
```
rkhamis: Try not to commit debuggers

PR #98, opened 2016-03-11
Somerefactoring

despiegk:
@zaibon 
also action force has been fixed


despiegk:
IMPORTANT we need to do the same replace in portal REPO !!!

die & showout !!!
are there other repo's?

PR #99, opened 2016-03-13
Docker host

zaibon:


lib/JumpScale/sal/docker/Docker.py
```diff
@@ -31,10 +31,14 @@ def isWeaveEnabled(self):
...
-        self.client = docker.Client(base_url=url)
-        self.docker_host = {'host': address, 'port': port}
+    def connectRemoteTCP(self, base_url):
+        self.base_url = base_url
+        self.client = docker.Client(base_url=base_url)
+
+    @property
+    def docker_host(self):
+        u = parse.urlparse(self.base_url)
+        return u.hostname
```
grimpy: Maybe this should throw erro in case of unix socket or return localhost?

PR #108, opened 2016-03-20
Influxdumper multipule ports

rkhamis:
I'm not 100% in love with this approach.
As it is, we scan for any host in the network that has any of these listed ports... we're making a big assumption that no conflicts can occur where another service is using the same port.

@zaibon what do you think?


muhamadazmy:
@rkhamis I also hate the scan approach, not only for this reason but also for the amount of time it takes to scan the entire IP range. If your IP range is wide the scan can take forever.

I suggest we totally modify this to our current use case, to only scan local host for the configured ports

lib/JumpScale/tools/aggregator/NetworkScanner.py
```diff
@@ -3,25 +3,25 @@
...
         """nmap -n --disable-arp-ping -send-ip -Pn -sS -p22 -oG - 172.17.0.1/24"""
 
-        cmd = self.COMMAND.format(port=self._port, cidr=self._cidr)
+        cmd = self.COMMAND.format(ports=self._ports, cidr=self._cidr)
         code, output = j.sal.process.execute(cmd, outputToStdout=False, die=False)
         if code != 0:
             raise RuntimeError('nmap scan failed')
 
-        regex = re.compile(r'^Host: ([^\(]+) .+Ports: %d/open' % self._port, re.MULTILINE)
+        regex = re.compile(r'^Host: ([^\(]+) .+Ports: %s/open' % self._ports, re.MULTILINE)
```
muhamadazmy: This scan method now is wrong

Before, the idea was to return a list of ALL the hosts that have port 7777 open. If need to change this to take multiple ports to test, the returned list should have different structure. 
Like
```python
hosts = {
  'host1': [port1, port2],
  'host2': [port2, port3]
}
```

Otherwise, I will still don't know what ports are open

muhamadazmy:
@rkhamis @zaibon 

The idea of the network-scanner was to find out which hosts are listening on a given port inside a certain IP range (defined by the cidr)

If we need to change this to accept multiple ports, we need to make sure of the following
1- nmap actually work with the given new parameters (multiple ports) and it lists out the hosts and the detected open ports in a parsable format)
2- return the right structure as per my comment.


IMHO: we can totally drop the network scanner in this new implementation. Since you already gonna scan localhost only, and you already know the list of open ports. we can just start the dumper with that information.
This will also fix an issue if one of the redis instances where down when we started the dumper and then it come back up, the dumper will just start pulling data from it. With the network scanner approach there is know why to know that the instance is alive again

zaibon:
I think now that we have the code for it. we should keep the ability to scan a range. 
But if we specify a host and a list of port, we should skip scanning and only use the host and ports given

grimpy:
Add a flag to bypass the scanner

grimpy:
After looking at the code I dont think we need this flag after all if we add a host with cidr / 32 scanning will take no time at all and result will be as expected.

zaibon:
@muhamadazmy  is the solution proposed by Jo ok for you ?
I'll let you review this and merge when you're done.

lib/JumpScale/tools/aggregator/Dumper.py
```diff
@@ -35,15 +35,15 @@ def cidr(self):
...
+    def ports(self):
+        return self._ports
 
     @property
     def candidates(self):
         return self._candidates
 
     def _process(self, ip, queue):
-        redis = j.clients.redis.getRedisClient(ip, self.port)
+        redis = j.clients.redis.getRedisClient(ip, self.ports)
```
muhamadazmy: I am sure this doesn't work! have you even tested this code ?

```python
#In js shell
j.clients.redis.getRedisClient?
Signature: j.clients.redis.getRedisClient(ipaddr, port, password='', fromcache=True)
```

grimpy:
Output of test now yields
```
[U influxdumper_multipule_ports][aggregator] jspython AggregatorClientTest.py
[Sun03 11:21] -        AggregatorClientTest.py:90   - INFO     - Injecting 100000 points for minute 0
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 40% points for minute 0
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 80% points for minute 0
[Sun03 11:21] -        AggregatorClientTest.py:103  - INFO     - Finished 100000 points for minute 0
[Sun03 11:21] -        AggregatorClientTest.py:90   - INFO     - Injecting 100000 points for minute 1
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 39% points for minute 1
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 80% points for minute 1
[Sun03 11:21] -        AggregatorClientTest.py:103  - INFO     - Finished 100000 points for minute 1
[Sun03 11:21] -        AggregatorClientTest.py:90   - INFO     - Injecting 100000 points for minute 2
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 38% points for minute 2
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 78% points for minute 2
[Sun03 11:21] -        AggregatorClientTest.py:103  - INFO     - Finished 100000 points for minute 2
[Sun03 11:21] -        AggregatorClientTest.py:90   - INFO     - Injecting 100000 points for minute 3
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 40% points for minute 3
[Sun03 11:21] -        AggregatorClientTest.py:100  - INFO     - Injected 82% points for minute 3
[Sun03 11:21] -        AggregatorClientTest.py:103  - INFO     - Finished 100000 points for minute 3
[Sun03 11:21] -        AggregatorClientTest.py:90   - INFO     - Injecting 100000 points for minute 4
[Sun03 11:22] -        AggregatorClientTest.py:100  - INFO     - Injected 40% points for minute 4
[Sun03 11:22] -        AggregatorClientTest.py:100  - INFO     - Injected 80% points for minute 4
[Sun03 11:22] -        AggregatorClientTest.py:103  - INFO     - Finished 100000 points for minute 4
[Sun03 11:22] - ...umpScale/sal/process/SystemProcess.py:1318 - INFO     - exec:which nmap
[Sun03 11:22] - ...umpScale/sal/process/SystemProcess.py:1364 - INFO     - system.process.execute [which nmap]
[Sun03 11:22] - ...umpScale/sal/process/SystemProcess.py:1318 - INFO     - exec:nmap -n --disable-arp-ping -send-ip -Pn -sS -p6379 -oX - 127.0.0.1
[Sun03 11:22] - ...umpScale/sal/process/SystemProcess.py:1364 - INFO     - system.process.execute [nmap -n --disable-arp-ping -send-ip -Pn -sS -p6379 -oX - 127.0.0.1]
```

lib/JumpScale/tools/aggregator/AggregatorClientTest.py
```diff
@@ -81,6 +87,8 @@ def statstest(self, aggregator, influxdb, points=1000000, minutes=5):
             start_time = time.time()
             totals = 0
             max = 0
+            self.logger.info("Injecting %s points for minute %s", points, minute)
+            start = time.time()
```
muhamadazmy: why you didn't remove 'start_time' created at line 87 since you have this now?

grimpy: Variables have a totaly diffrent usecase. One is for logging every 5sec. the other one for calculating the rate.

lib/JumpScale/tools/aggregator/AggregatorClientTest.py
```diff
@@ -88,7 +96,11 @@ def statstest(self, aggregator, influxdb, points=1000000, minutes=5):
                 totals += val
                 # always use the `now` as time stamp so we have control which values are lying in this minute
                 aggregator.measure(key, "random", "mode:test", val, timestamp=stamp)
+                if time.time() - start > 5:
```
muhamadazmy: I guess you missed something of the logic here. I don't need to run this loop for 5 min or whatever time is given. I just simulate running for 5 min but I can go must faster, incrementing the clock as fast as I want.
Influxdb doesn't care if you give timestamp in the future.

grimpy: This is just for logging sake to see a log message every 5 seconds to see the progress.

despiegk:
Let's talk about this

PR #115, opened 2016-03-23
Services has an executor property

rkhamis:
Service would check its parents until it finds a getExecutor defined.
If none of its parents are have it implemented, will use a local executor


zaibon:
Looks good.

PR #127, opened 2016-04-03
Checking if jumpscale is installed

NourhanOmar:


lib/JumpScale/tools/cuisine/CuisineBuilder.py
```diff
@@ -13,11 +13,13 @@ class CuisineBuilder(object):
     def __init__(self, executor, cuisine):
         self.executor = executor
         self.cuisine = cuisine
+    def jumpscale_installed(self):
+        return self.cuisine.executor.execute('which js')
```
rkhamis: this will either give a RuntimeError or return the bin path.
You should check the exit code after ensuring it doesn't die (die=False) and go from there.

lib/JumpScale/tools/cuisine/CuisineBuilder.py
```diff
@@ -13,11 +13,22 @@ class CuisineBuilder(object):
     def __init__(self, executor, cuisine):
         self.executor = executor
         self.cuisine = cuisine
+    def jumpscale_installed(self, die=False):
+        try:
+            result = self.cuisine.executor.execute('which js')
```
rkhamis: No, no. Just use `rc, out = self.cuisine.executor.execute('which js', die=False)`
rc would be 0 for a successful execution

PR #129, opened 2016-04-04
bug fix in jumpcale.log creation,

abdulrahmantkhalifa:
i was not sure where to put it, i  can put the condidtion in __fileRotateHandler method , but i dont see the difference


lib/JumpScale/core/logging/LoggerFactory.py
```diff
@@ -30,7 +30,8 @@ class LoggerFactory:
     def __init__(self):
         self.__jslocation__ = "j.logger"
         self.root_logger_name = 'j'
-
+        if not os.path.exists("/optvar/log/"):
+            os.mkdir("/optvar/log/")
```
grimpy: Doubt we can get away with harcoding these paths lile this in this part of the code.

zaibon:
I would indeed put it in the __fileRotateHandler method, cause this directory is only needed for this specific handler. let's try to keep codes that deal with the same problem close.

lib/JumpScale/core/logging/LoggerFactory.py
zaibon: Well there is an issue with that. 
Because the logger is loaded during Jumpscale bootstrap, the j.dirs is not yet loaded so we can't read it at that moment. 
Still need to figure out how to do that.

PR #137, opened 2016-04-06
Removing j.application from filter, fixing a bug in logger factory an…

NourhanOmar:
…d creating a logger in the run method of cuisinecore


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -957,6 +957,7 @@ def run(self,cmd,die=True,debug=None,checkok=False,showout=True,profile=False,re
         @param profile, execute the bash profile first
         """
         # print (cmd)
+        j.logger.get('j.tools.cuisine.core', enable_only_me=True)
```
rkhamis: why not in init?

PR #141, opened 2016-04-07
Optmise message handling in telegrambot

abdulrahmantkhalifa:
tested


zaibon:
I won't accept this. 
It binds the Asker class to the implementation of the bot that use it.

PR #143, opened 2016-04-10
Install tools fix

abdulrahmantkhalifa:


lib/JumpScale/clients/git/GitFactory.py
```diff
@@ -116,3 +116,191 @@ def _getRepos(codeDir, account=None, name=None): # NOQA
...
+        """
+
+        if url == "":
+            if dest == None:
+                raise RuntimeError("dest cannot be None (url is also '')")
+            if not j.sal.fs.exists(dest):
+                raise RuntimeError(
+                    "Could not find git repo path:%s, url was not specified so git destination needs to be specified." % (dest))
+
+        if login == None and url.find("github.com/") != -1:
```
rkhamis: I feel if we're explicitly checking for "github", this logic is better moved to the github factory

lib/JumpScale/clients/git/GitFactory.py
```diff
@@ -116,3 +116,191 @@ def _getRepos(codeDir, account=None, name=None): # NOQA
...
+            # can see if there if login & passwd in OS env
+            # if yes fill it in
+            if "GITHUBUSER" in os.environ:
```
rkhamis: env variables will almost never be set when you're using this library

lib/JumpScale/clients/git/GitFactory.py
```diff
@@ -116,3 +116,191 @@ def _getRepos(codeDir, account=None, name=None): # NOQA
...
+        protocol, repository_host, repository_account, repository_name, repository_url = j.do.rewriteGitRepoUrl(
+            url=url, login=login, passwd=passwd, ssh=ssh)
+
+        repository_type = repository_host.split(
+            '.')[0] if '.' in repository_host else repository_host
+
+        if not dest:
+            if codeDir == None:
+                if executor == None:
+                    codeDir = j.dirs.codeDir
```
rkhamis: if executor is none (make the executor an arg of init of the factory), it should get a local executor

lib/JumpScale/clients/git/GitFactory.py
```diff
@@ -116,3 +116,191 @@ def _getRepos(codeDir, account=None, name=None): # NOQA
...
+        if ssh == "first":
+            try:
+                return self.pullGitRepo(url, dest, login, passwd, depth, ignorelocalchanges, reset, branch, revision, True, executor)
+            except Exception as e:
+                return self.pullGitRepo(url, dest, login, passwd, depth, ignorelocalchanges, reset, branch, revision, False, executor)
+            raise RuntimeError(
+                "Could not checkout, needs to be with ssh or without.")
+
+        if executor is None:
+            executor = self
```
rkhamis: definitely not tested. This `self` will break everything. See earlier comment about how to instantiate the client and how to handle the executor in init

lib/JumpScale/clients/github/Github.py
```diff
@@ -34,3 +36,187 @@ def identify(self):
...
+
+        protocol, repository_host, repository_account, repository_name, repository_url = j.do.rewriteGitRepoUrl(
+            url=url, login=login, passwd=passwd, ssh=ssh)
+
+        repository_type = repository_host.split(
+            '.')[0] if '.' in repository_host else repository_host
+
+        if not dest:
+            if codeDir == None:
+                if self.executor == None:
```
rkhamis: self.executor could never be `None` according to the factory

PR #144, opened 2016-04-10
added the libvirt method not tested

abdulrahmantkhalifa:


zaibon:
Shouldn't this be also a cuisine app ?
Installation probably also need libvirt on the host machine a start/stop method.

rkhamis:
What's the status on this?
And if it's not tested, why is there a PR?

PR #149, opened 2016-04-11
Warn if templates don't have matching interfaces

rkhamis:


zaibon:
@despiegk What do you think of that ?

despiegk:
I don't understand, lets talk
not sure what you try to accomplish

rkhamis:
@despiegk it just checks that all service templates with the same role have matching methods defined.

The idea here is that service templates with the same role are interchangeable and as such should expose the same functionality.

PR #151, opened 2016-04-12
grafana client: add ping method

zaibon:


lib/JumpScale/clients/grafana/Grafana.py
```diff
@@ -29,6 +29,14 @@ def __init__(self, url, username, password):
         self._url = url
         self.setAuth(username, password)
 
+    def ping(self):
+        url = os.path.join(self._url, 'api/search/')
+        try:
+            self._session.get(url).json()
+            return True
+        except:
```
grimpy: Very broad exception? Probably when you have wrong authentication this will give a false negative.

lib/JumpScale/clients/grafana/Grafana.py
```diff
@@ -29,6 +29,14 @@ def __init__(self, url, username, password):
         self._url = url
         self.setAuth(username, password)
 
+    def ping(self):
+        url = os.path.join(self._url, 'api/search/')
```
muhamadazmy: isn't there a lighter weight endpoint to call instead of search ?

lib/JumpScale/clients/grafana/Grafana.py
zaibon: if auth is wrong, then we can't use it anyway right ? So it's still correct to return false no ?

lib/JumpScale/clients/grafana/Grafana.py
zaibon: good point, I'll look at a light endpoint in the API

muhamadazmy: may be `/api/org/` is lighter/faster endpoint to check availability with.

lib/JumpScale/clients/grafana/Grafana.py
grimpy: Actually apparently this will just return a message saying something is wrong with the username/password.
Ping will succeed. For authtest you would have to validate the status_code (401)

zaibon: So maybe I could just alias ping with isAuthenticated ? or just drop ping and use isAuthenticated when I want to test connectivity to grafana

PR #156, opened 2016-04-12
Adds fix to monoengine version

ashraffouda:


lib/JumpScale/tools/cuisine/CuisinePIP.py
```diff
@@ -39,7 +39,8 @@ def install(self,package=None,upgrade=False):
                 
         if package == "pymongo":
             package = "pymongo==3.2.1" 
-
+        if package == "mongoengine":
```
rkhamis: I think it would be cleaner to set the versions wherever they're installed and keep pip unaware of what package is being installed

PR #160, opened 2016-04-12
added binDir to profile_js on cuisine.run

abdulrahmantkhalifa:
the path /opt/jumpscale8/bin is now added when trying
to run something through cuisine with profile= True


lib/JumpScale/tools/bash/Bash.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, content):
         else:
             self._path = set()
             self._path.add('${PATH}')
+            self._path.add(j.dirs.binDir)
```
rkhamis: You're adding the local bin dir, not the remote's.

lib/JumpScale/tools/bash/Bash.py
```diff
@@ -204,6 +205,7 @@ def profile(self):
             if self.cuisine.core.file_exists(self.profilePath):
                 content = self.cuisine.core.file_read(self.profilePath)
             self._profile = Profile(content)
+            self.cuisine.core.file_write(self.profilePath, self._profile.dump())
```
rkhamis: You don't need to write here. Why? You'll be writing the file each time the property is loaded
What's the use-case of this?

lib/JumpScale/tools/bash/Bash.py
```diff
@@ -13,7 +13,7 @@ class Profile(object):
     env_pattern = re.compile(r'^([^=\n]+)="([^"\n]+)"$', re.MULTILINE)
     include_pattern = re.compile(r'^source (.*)$', re.MULTILINE)
 
-    def __init__(self, content):
+    def __init__(self, content, binDir):
```
grimpy: default binDir to None?

PR #165, opened 2016-04-14
Adds uninstall method to ays to remvoe ssh-key

### LINKED ISSUE #138: When too many ssh keys loaded in ssh-agent. Connection is refused from most of SSH servers.
Most of the ssh server won't accept more then x keys when trying to connect.
If the valid key is loaded in ssh-agent but not tried in time. connection will fail

ashraffouda:
Fixes #138

### DIFF
--- a/install/InstallTools.py
+++ b/install/InstallTools.py
@@ -1410,7 +1410,7 @@ class InstallTools():
             return None
 
         for line in out.splitlines():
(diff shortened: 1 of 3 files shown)


despiegk:
OK

despiegk:
bugs
- [js8:138:title] (https://github.com/Jumpscale/jumpscale_core8/issues/138)

PR #168, opened 2016-04-14
Cuisine builder fix

abdulrahmantkhalifa:
Hadnling the exit code after it occurs from the tmux lsw, command will halt execution due to the way docker files RUN commands , so i have to handle it before that i was gonna write a script for it but @grimpy suggested to handle the other error case, where tmux is not installed , in a condition before the command and set the exit code manually to acoid this situation.

aslo,  if possible do not  delete branch,  might need it for further bugs.

### DIFF
--- a/lib/JumpScale/tools/cuisine/CuisineCore.py
+++ b/lib/JumpScale/tools/cuisine/CuisineCore.py
@@ -1155,9 +1155,14 @@ class CuisineCore():
     #####################SYSTEM IDENTIFICATION
     @property
     def isDocker(self):
-        docker = self.run('mount | grep hostname > /dev/null', die = False)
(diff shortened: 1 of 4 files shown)


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1155,9 +1155,15 @@ def execute_jumpscript(self,script):
     #####################SYSTEM IDENTIFICATION
     @property
     def isDocker(self):
-        docker = self.run('mount | grep hostname > /dev/null', die = False)
-        return not docker[0]
+        import ipdb;ipdb.set_trace()
```
rkhamis: debugger?

PR #172, opened 2016-04-14
better implementation  in isDocker and isLxc mthods

abdulrahmantkhalifa:
use re lib in python instead of executionusing the bash

### DIFF
--- a/lib/JumpScale/tools/cuisine/CuisineCore.py
+++ b/lib/JumpScale/tools/cuisine/CuisineCore.py
@@ -200,6 +200,7 @@ class CuisineCore():
         self.cd="/"
         self._dirs={}
         self.sudomode = False
+        self.__cgroup = None
 
         self.executor = executor
         self.cuisine=cuisine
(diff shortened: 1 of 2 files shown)


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1155,12 +1155,14 @@ def execute_jumpscript(self,script):
     #####################SYSTEM IDENTIFICATION
     @property
     def isDocker(self):
-        self._isDocker = not int(self.run('cat  /proc/1/cgroup | grep "docker" >/dev/null ; echo $?', die=False)[1])
+        self._cgroup = self.file_read("/proc/1/cgroup")
```
grimpy: why store this in an instance attrivbute? would make sense it you reuse the same one for both isDocker and isLXC but your dont?

lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1155,12 +1155,14 @@ def execute_jumpscript(self,script):
...
+        self._isDocker = True if re.search("docker", self._cgroup) else False
         return self._isDocker
 
     @property
     def isLxc(self):
-        self._isLxc = not int(self.run('cat  /proc/1/cgroup | grep "lxc" >/dev/null ; echo $?', die=False)[1])
+        self._cgroup = self.file_read("/proc/1/cgroup")
+        self._isLxc = True if re.search("lxc", self._cgroup) else False
```
grimpy: since you are doing basic string matching you can use find instead

Issue #190, opened 2016-04-20
fileGetUncommentedContents doesn't work for comments that start with spaces


xmonader:
```python
    def fileGetUncommentedContents(self, filename):
        """Read a file and get uncommented contents of that file
        @param filename: string (filename to open for reading )
        @rtype: list of lines of uncommented file contents
        """
        if filename is None:
            raise TypeError('File name is None in system.fs.fileGetContents')
        self.logger.debug('Opened file %s for reading'% filename)
        # self.logger.debug('Reading file %s'% filename,9)
        with open(filename) as fp:
            data = fp.readlines()
        uncommented = list()
        for line in data:
            if not line.startswith('#') and not line.startswith('\n'):
                line = line.replace('\n', '')
                uncommented.append(line)
        self.logger.debug('File %s is closed after reading'%filename)
        return uncommented

```

Suggested fix: 
```python

    def fileGetUncommentedContents(self, filename):
        """Read a file and get uncommented contents of that file
        @param filename: string (filename to open for reading )
        @rtype: list of lines of uncommented file contents
        """
        if filename is None:
            raise TypeError('File name is None in system.fs.fileGetContents')
        self.logger.debug('Opened file %s for reading'% filename)
        # self.logger.debug('Reading file %s'% filename,9)
        with open(filename) as fp:
            data = fp.readlines()
        uncommented = list()
        for line in data:
            line = line.strip() #fixes lines start with spaces.
            if not line.strip().startswith('#') and not line.startswith('\n'):  #keeping unindented lines
                line = line.replace('\n', '')
                uncommented.append(line)
        self.logger.debug('File %s is closed after reading'%filename)
        return uncommented

```

Issue #187, opened 2016-04-20
sal/fs md5sum fails


xmonader:
salf/fs/md5sum and getFolderMD5sum fail because

```

RuntimeError: ERROR: Failed to get the hex digest of the file ['/home/hello.py']in system.fs.md5sum. Error: Unicode-objects must be encoded before hashing ((type:runtime.error)

```

```python
    def md5sum(self, filename):
        """Return the hex digest of a file without loading it all into memory
        @param filename: string (filename to get the hex digest of it) or list of files
        @rtype: md5 of the file
        """
        self.logger.debug('Get the hex digest of file %s without loading it all into memory'%filename)
        if filename is None:
            raise('File name is None in system.fs.md5sum')
        if not isinstance(filename, list):
            filename = [filename]
        digest = hashlib.md5()
        try:
            for filepath in filename:
                with open(filepath) as fh:
                    while True:
                        buf = fh.read(4096)
                        if buf == "":
                            break
                        digest.update(buf)
            return digest.hexdigest()
        except Exception as e:
            raise j.exceptions.RuntimeError("Failed to get the hex digest of the file %sin system.fs.md5sum. Error: %s"  % (filename,str(e)))

```
Suggested fix:
1- Encode buf as utf-8
```python

    def md5sum(self, filename):
        """Return the hex digest of a file without loading it all into memory
        @param filename: string (filename to get the hex digest of it) or list of files
        @rtype: md5 of the file
        """
        self.logger.debug('Get the hex digest of file %s without loading it all into memory'%filename)
        if filename is None:
            raise('File name is None in system.fs.md5sum')
        if not isinstance(filename, list):
            filename = [filename]
        digest = hashlib.md5()
        try:
            for filepath in filename:
                with open(filepath) as fh:
                    while True:
                        buf = fh.read(4096)
                        if buf == "":
                            break
                        buf=buf.encode("utf-8")
                        digest.update(buf)
            return digest.hexdigest()
        except Exception as e:
            raise j.exceptions.RuntimeError("Failed to get the hex digest of the file %sin system.fs.md5sum. Error: %s"  % (filename,str(e)))

```
2- Just open the file in 'rb' mode thanks to @mie00 

```python

    def md5sum(self, filename):
        """Return the hex digest of a file without loading it all into memory
        @param filename: string (filename to get the hex digest of it) or list of files
        @rtype: md5 of the file
        """
[...]

Issue #186, opened 2016-04-20
sal.ubuntu checkroot fails


xmonader:
```python
    @property
    def whoami(self):
        rc,result=self._local.execute("whoami")
        return result.strip()
    def checkroot(self):
        if self.whoami() != "root": #calling the property whoami...
            j.events.inputerror_critical("only support root")
```

suggested fixes:
1- remove @property decorator from the whoami definition
2- remove the self.whoami method call and access the whoami property

Isn't it supposed to return True if root and False if otherwise? it returns None if self.whoami == 'root'

abdulrahmantkhalifa:
different implementation of fixes choose reviewer chose one

PR #192, opened 2016-04-21
Salfixes

### LINKED ISSUE #182: sal fileGetTextContents
This method fails in the testing because mainly we open the file with the default 'ascii' encoding

```python
    def fileGetTextContents(self, filename):
        """Read a UTF-8 file and get contents of that file. Takes care of the [BOM](http://en.wikipedia.org/wiki/Byte_order_mark)
        @param filename: string (filename to open for reading)
        @rtype: string representing the file contents
        """
        if filename is None:
            raise TypeError('File name is None in system.fs.fileGetTextContents')
        with open(filename) as f:
            s = f.read()

        boms = [codecs.BOM_UTF8]
        for bom in boms:  # we can add more BOMs later:
            if s.startswith(bom.decode()):
                s = s.replace(bom.decode(), '', 1)
                break
        return s


```
Suggested fixes
1-

```python
    def fileGetTextContents(self, filename):
        """Read a UTF-8 file and get contents of that file. Takes care of the [BOM](http://en.wikipedia.org/wiki/Byte_order_mark)
        @param filename: string (filename to open for reading)
        @rtype: string representing the file contents
        """
        if filename is None:
            raise TypeError('File name is None in system.fs.fileGetTextContents')
        with open(filename, encoding='utf8') as f: #enforce utf8 encoding
            s = f.read()

        boms = [codecs.BOM_UTF8]
        for bom in boms:  # we can add more BOMs later:
            if s.startswith(bom.decode()):
                s = s.replace(bom.decode(), '', 1)
                break
        return s


```
2- using utf_8_sig from the beginning in the implementation as python recommends this way
https://docs.python.org/3/library/codecs.html#module-encodings.utf_8_sig
something like that maybe
```python
def stripbom(bytes):
    return bytes.decode(encoding='utf_8_sig')
```

xmonader:
Fixes  #182 #186 #187 #190

### DIFF
--- a/lib/JumpScale/sal/fs/SystemFS.py
+++ b/lib/JumpScale/sal/fs/SystemFS.py
@@ -1674,7 +1674,8 @@ class SystemFS(SALObject):
             dir = dir or j.dirs.tmpDir
             return tempfile.mktemp('', prefix, dir)
 
-    def isAsciiFile(self, filename):
+    def isAsciiFile(self, filename, checksize=4096):
+        #TODO: let's talk about checksize feature.
         try:
             with open(filename, encoding='ascii') as f:
                 f.read()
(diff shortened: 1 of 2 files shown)


lib/JumpScale/sal/fs/SystemFS.py
```diff
@@ -1343,20 +1343,22 @@ def fileGetUncommentedContents(self, filename):
             data = fp.readlines()
         uncommented = list()
         for line in data:
+            line = line.strip() #fixes lines start with spaces.
```
rkhamis: doesn't that mean you're returning unindented lines? Not the same content

xmonader: I've no idea why the indentations were so messed up. I think they are fixed now :) Thanks

PR #194, opened 2016-04-21
Cuisine libvirt installer

abdulrahmantkhalifa:
tested

### DIFF
--- a/lib/JumpScale/tools/cuisine/CuisineInstaller.py
+++ b/lib/JumpScale/tools/cuisine/CuisineInstaller.py
@@ -133,12 +133,9 @@ class CuisineInstaller(object):
         """
         do not use in containers or VMs only actual machines @todo not tested 
         """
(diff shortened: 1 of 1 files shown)


lib/JumpScale/tools/cuisine/CuisineInstaller.py
```diff
@@ -129,6 +129,14 @@ def jumpscale8(self, rw=False, reset=False):
         print ("* re-login into your shell to have access to js, because otherwise the env arguments are not set properly.")
 
     @actionrun(action=True)
+    def libvirt(self):
+        """
+        do not use in containers or VMs only actual machines @todo not tested 
+        """
```
rkhamis: Is it still not tested?

abdulrahmantkhalifa: tested old code my bad

PR #198, opened 2016-04-26
fix for j.do.gitPullRepo when taking branch

abdulrahmantkhalifa:
now does not always default to master , also removed an extra condition
at the end commented only as not certain if targeted at specific case

### DIFF
--- a/install/InstallTools.py
+++ b/install/InstallTools.py
@@ -1749,7 +1749,6 @@ class InstallTools():
         @param ssh ==True means will checkout ssh
         @param ssh =="first" means will checkout sss first if that does not work will go to http
         """
-
         if ssh=="first":
             try:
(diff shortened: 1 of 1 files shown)


install/InstallTools.py
```diff
@@ -1819,8 +1818,8 @@ def pullGitRepo(self,url="",dest=None,login=None,passwd=None,depth=1,ignorelocal
 
             print(cmd)
 
-            if depth!=None:
-                cmd+=" --depth %s"%depth
```
muhamadazmy: I guess should be gone as well but fix the condition at line 1807 to be 

```python
             if depth:
                  extra = "--depth=%s" % depth
```

So only depth is added if depth is not a False value

PR #199, opened 2016-04-26
Email sal

muhamadazmy:


lib/JumpScale/tools/email/Email.py
```diff
@@ -0,0 +1,69 @@
...
+
+        :return: list
+        """
+
+        messages=[]
+        keys=self._get_lastest_keys()
+        for k in keys:
+            #get k from the hashset
+            m=Message(j.core.db.hmget('mails', key))
+            m.set_key(k)
```
muhamadazmy: I think the key must be passed in the Message __init__ . it's mandatory for the Message operation so you should not have a Message without a key

lib/JumpScale/tools/email/Message.py
```diff
@@ -0,0 +1,112 @@
...
+    def set_key(self, key):
+        self._key = key
+
+    def _cache(self):
+        if self._key:
+            self._cache = True
+            self.update(util.get_json_msg(self._key))
+
+    ##THESE ARE LOADED LAZELY.
+    def lazy(self, f):
```
muhamadazmy: This can't be used as a decorator in this use case.

lib/JumpScale/tools/email/Message.py
```diff
@@ -0,0 +1,112 @@
...
+        super().__init__(*args, **kwargs)
+        self._key=None
+        self._cached = False
+
+    def set_key(self, key):
+        self._key = key
+
+    def _cache(self):
+        if self._key:
+            self._cache = True
```
muhamadazmy: set this after the loading of the file, so if the file loading crashed, you don't ruin the message state

muhamadazmy:
@xmonader can you please make sure you follow pep8 rules. for spacing and new lines.
You can set your editor to do the pep8 validation automatically. after that we can merge.

despiegk:
to which story or task does this pull request relate too?
please put link

PR #205, opened 2016-04-28
Improving skydns and adding mail (MX and TXT) support

maxux:


despiegk:
do we still use this?
thought we were going to go to other dns server?

zaibon:
see https://github.com/gig-projects/org_development/issues/5#issuecomment-212278550

despiegk:
see story, indeed, security is not an issue

despiegk:
please merge into master so we don't loose the changes

Issue #174, opened 2016-04-17
Rename OurExceptions to something more meaningful


muhamadazmy:
I suggest to 'JSException'

despiegk:
don't understand the purpose and also not whatwe are trying to do

Issue #148, opened 2016-04-11
Use `j.exceptions` instead of `j.errorconditionhandler` everywhere


rkhamis:

mie00:
Error handling in jumpscale

This is a proposal regarding exception in jumpscale:

1. Exceptions should be a subclass of our base exception class, they should have no side effects so they'll just contain data.

2. Exceptions should always be raised.

3. Warnings should has two parent classes oue base exception class and builtin `Warning`.

4. They should be used only to dump a warning to the stderr. they can be filtered or be raised an exception and they should follow [this](https://docs.python.org/3.3/library/warnings.html#module-warnings).

5. Logging of the exception and storing it in db shouldn't be the reponsiblity of the place where the exception occurs, instead it should be the responsibility of who handles it. ha can always find another way or log the exception and store it in the database.

6. If the exception is not handled it will propagate until it reaches our exception hook where he can always log the exception and store it in the database.

7. We can use j.logger.exception under an except statement to print the exception and it's traceback. see [here](https://docs.python.org/2/library/logging.html#logging.Logger.exception).

8. Exceptions should be serialized and stored in database.

9. Exceptions can have a serialize method to add custom attributes.

10. There is a namespace (eg. a class) to contain helpers regarding handling exceptions.

11. In this namespace there is a method to serialize exception objects and it should respect Exception.serialize method if it exists, and the output of this method is what will be stored in the database.

12. In this namespace there is a method to store the serialized object in the database to be used afterwords.

13. The two previous methods hould only be used in except statement where it gets the exception object from the `__context__` variable.

14. The exception class name should be part of what it shtored in the db.

15. The serialized data can be stored later on mongodb, so we can normally query them.

muhamadazmy:
Proposal looks fine and matches what we discussed.

despiegk:
you mention
5. Logging of the exception and storing it in db shouldn't be the reponsiblity of the place where the exception occurs, instead it should be the responsibility of who handles it. ha can always find another way or log the exception and store it in the database.

could be dangerous, what if people do not process it?
with db do you mean redis or something else?

WANT TO MAKE SURE
- we never lose track of an error happening

mie00:
Handling an exception using except statement gives you full responsibility of the exception, that includes logging and storing or even reraising it depending on the situation, so normally none of that will happen automatically if you catch it. 

If the exception is not handled manually , it will be handled using a custom excepthook, and it will always log and store it. 

[...]

PR #207, opened 2016-04-28
Exception handling

mie00:
This is the basic implementation regarding this issue #148 , and that #174.

We still need to migrate from `j.errorconditionhandler` `j.events` inside the code itself.

Example usages:
1. `test1.py`

```
from JumpScale import j

raise j.exceptions.RuntimeError('error 1')
```

will print

```
[Thu28 16:38] - .../core/errorhandling/ExceptionUtils.py:61   - ERROR    - Traceback (most recent call last):
  File "test1.py", line 3, in <module>
    raise j.exceptions.RuntimeError('error 1')
core.errorhandling.Exceptions.RuntimeError: ERROR: error 1 ((type:runtime.error))
```
1. `test2.py`

```
from JumpScale import j

logger = j.logger.get('test2')

try:
    raise j.exceptions.RuntimeError('error 2')
except:
    logger.error_tb()

this_will_raise_a_name_exception
```

will print

```
[Thu28 16:40] -                       test2.py:8    - ERROR    - Traceback (most recent call last):
  File "test2.py", line 6, in <module>
    raise j.exceptions.RuntimeError('error 2')
core.errorhandling.Exceptions.RuntimeError: ERROR: error 2 ((type:runtime.error))

[Thu28 16:40] - .../core/errorhandling/ExceptionUtils.py:61   - ERROR    - Traceback (most recent call last):
  File "test2.py", line 11, in <module>
    this_will_raise_a_name_exception
NameError: name 'this_will_raise_a_name_exception' is not defined
```
1. `test3.py`

```
from JumpScale import j

logger = j.logger.get('test3')

logger.warn_tb(j.exceptions.RuntimeError, 'error 3')
```


apps/agentcontroller/jumpscripts/jumpscale/healthcheck_db.py
```diff
@@ -20,17 +20,18 @@
 
 
 def action():
+    logger = j.logger.get('healthcheck_db')
     osiscl = j.clients.osis.getByInstance('main')
     status = osiscl.getStatus()
     results = list()
     if status['mongodb'] == False :
-        j.errorconditionhandler.raiseOperationalCritical('mongodb status -> halted', 'monitoring', die=False)
+        logger.warn_tb(j.exceptions.OPERATIONS, 'mongodb status -> halted')
```
muhamadazmy: Is this gonna raise an exception? It's not clear from the name that this one gonna raise an exception and kill the script. If it's raising an exception should be renamed to 'raiseWarn' or 'raise_warn'

apps/agentcontroller/jumpscripts/jumpscale/healthcheck_disks.py
```diff
@@ -50,10 +51,10 @@ def disktoStr(disk):
         if disk.free and disk.size:
             freepercent = (disk.free / float(disk.size)) * 100
             if checkusage and (freepercent < 10):
-                j.errorconditionhandler.raiseOperationalWarning(result['message'], 'monitoring')
+                logger.warn_tb(j.exceptions.OPERATIONS, result['message'])
                 result['state'] = 'WARNING'
             if checkusage and (freepercent < 5):
-                j.errorconditionhandler.raiseOperationalCritical(result['message'], 'monitoring', die=False)
+                logger.warn_tb(j.exceptions.OPERATIONS, result['message'])
```
muhamadazmy: I believe critical level is very important to retain. Critical means system is in danger and needs immediate interfere by the operators to fix it. Unlike warnings which are usually self recoverable.

lib/JumpScale/core/logging/JSLogger.py
```diff
@@ -1,50 +1,35 @@
...
-        """
-        Log 'msg % args' with severity 'ERROR'.
-
-        To pass exception information, use the keyword argument exc_info with
-        a true value, e.g.
-
-        logger.error("Houston, we have a %s", "major problem", exc_info=1)
-
-        """
+    def error_tb(self, ttype=None, exceptionObject=None, tb=None, store=True):
```
muhamadazmy: Not sure we still need to have the `error` and `warn` methods from the base class. May be it's better to hide them entirely. So rename this to `error` and `warn` instead of `error_tb` and `warn_tb`. Please confirm with @rkhamis and @zaibon

apps/agentcontroller/jumpscripts/jumpscale/healthcheck_jsprocess_status.py
```diff
@@ -32,7 +33,7 @@ def action():
         result['category'] = 'AYS Process'
         if not ays.actions.check_up_local(ays, wait=False):
             message = "Process %s:%s:%s is halted" % (ays.domain, ays.name, ays.instance)
-            j.errorconditionhandler.raiseOperationalWarning(message, 'monitoring')
+            logger.warn_tb(j.exceptions.OPERATIONS, message)
```
muhamadazmy: shouldn't the first argument `j.exceptions.OPERATIONS` moved to the last argument and be defaulted to j.exceptions.OPERATIONS.

lib/JumpScale/tools/aggregator/eco.lua
```diff
@@ -1,70 +0,0 @@
---message, messagepub, str(level), type, tags, code, funcname, funcfilepath, backtrace, str(time)
```
muhamadazmy: Are you sure the aggregator loads the correct lua file after this one has been moved ?

apps/agentcontroller/jumpscripts/jumpscale/healthcheck_jsprocess_status.py
mie00: The advantage of this is that we can pass multiple arguments to the exception.

lib/JumpScale/tools/aggregator/eco.lua
mie00: It's done here https://github.com/Jumpscale/jumpscale_core8/pull/207/files/3298925eefb5d537a23ec2084131262b7d703ba7#diff-d85e182f984d10280150a11704f02070R19

apps/agentcontroller/jumpscripts/jumpscale/healthcheck_db.py
mie00: Previously if die=False was passed to `raiseOperationalCritical` it will not stop the execution. It will log the exception and store it and that is basically what `warn_tb` does.

lib/JumpScale/core/logging/JSLogger.py
mie00: Apparently warn, error methods are used in jumpscale and `error_tb`, `warn_tb` have different signature, so they cannot replace them.

rkhamis:
@muhamadazmy why has this never been merged?

rkhamis:
This is ridiculously out of date. Will close and remove branch.

PR #210, opened 2016-05-01
Ays8fix

despiegk:
still open, needs much more verification, its just to make it easy to see what is happening
https://github.com/gig-projects/org_development/issues/99


zaibon:
This start to drift far away from master, can we plan a merge ?

rkhamis:
Approach to consumption self-healing is through a method "consume" that contains the consumption code of another service. It will make self-healing a lot easier if AYS knows exactly what to run when a producer service changes.
For example, portal consumes mongo. In the portal actions, we should define a consume method that deals with the consumption of mongo
```
def consume(self):
    ....
    mongo = self.service.producers.get('mongo')
    portal_connection['mongo'] = monogo.hrd.get('port')

    # do portal configuration
    ....

```
that way, if the mongo instance hrd or the template hrd, the change is propagated to its consumers (in this case, the portal) and the consume runs again.
Currently, a change in AYS (whether in its actions, template hrd or instance hrd), sets its consumers' consume method to "CHANGED" as well.

Not all cases of consumption will need `consume` methods defined

For example, a node.docker consumes an app.docker service. Because it requires that docker be installed on a host before creating a docker.
This consumption is handled through the schema.hrd file and no further logic is needed.

PR #211, opened 2016-05-03
fix bug in installtools

abdulrahmantkhalifa:
executor takes an extra parameter which the type of executor doesnt
handle
relates to  :https://github.com/Jumpscale/home/issues/8


zaibon:
The bug was due to a poor implementation of the execute command in our SSHClient.
Commit f7d5f94571b00d3c457b07c789ec8b9f7f5e6064 improve the implementation and fix the issue.

Sandboxing works now also when using SSH executor.

abdulrahmantkhalifa:
tested it works

PR #215, opened 2016-05-10
Fix 180

### LINKED ISSUE #180: Install script tries to retrieve non-existing script
In the install script https://github.com/Jumpscale/jumpscale_core8/blob/master/install/install.sh#L88 the code tries to download a helper script that does not exist in the same location anymore in the repo.

file url in script: https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/master/install/web/InstallTools.py

file url in the repo: https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/master/install/InstallTools.py

Another the problem is that the install script does not fail when the command fails.

xmonader:
Fixes #180 : instead of downloading/executing InstallTools.py, we leave that to bootstrap.py

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -85,7 +85,7 @@ fi
 set -ex
 branch=${JSBRANCH-master}
 curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/install/web/bootstrap.py > $TMPDIR/bootstrap.py
-curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/install/InstallTools.py > $TMPDIR/InstallTools.py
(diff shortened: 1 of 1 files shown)


install/install.sh
```diff
@@ -85,7 +85,7 @@ fi
 set -ex
 branch=${JSBRANCH-master}
 curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/install/web/bootstrap.py > $TMPDIR/bootstrap.py
-curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/install/web/InstallTools.py > $TMPDIR/InstallTools.py
+#curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/install/InstallTools.py > $TMPDIR/InstallTools.py
```
rkhamis: Why not remove it altogether?

ahussein: I also do not understand why the code disable shell failure before the curl command

PR #217, opened 2016-05-11
Fs testing

rkhamis:


tests/sal/SystemFS.py
```diff
@@ -0,0 +1,331 @@
...
+        assert_equal(os.listdir(d), [])
+        assert_equal(fs.isEmptyDir(d), True)
+        os.rmdir(d)
+
+    def test_isAbsolute(self):
+        assert_equal(path.isabs(TestFS.FILE), fs.isAbsolute(TestFS.FILE))
+
+
+    def test_touch(self):
+        f=tmpify('testx1')
```
rkhamis: Try with invalid path. Assert will fail

tests/sal/SystemFS.py
```diff
@@ -0,0 +1,331 @@
+from JumpScale import j
```
rkhamis: Make sure tests include "unhappy path" aka invalid data or failure scenarios.y 
Validate behavior in both success and failure

Try special characters in paths?

rkhamis: Try with non-keyword arguments.
Try with more or less arguments.
Try with wrong type of arguments --> list instead of path

ahussein:
I will merge the PR but this is not how we will be writing unittests in the future. Our unittests should not write files the system and leave the system in a different state compared with the state before running the tests

PR #224, opened 2016-05-17
Issue 88

ashraffouda:


lib/JumpScale/clients/cockpit/CockpitBot.py
```diff
@@ -148,3 +156,46 @@ def run(self):
...
+            self._check_services(chat_id, username, get_status=True)
+        except Exception as e:
+            self.logger.error(e)
+            self.bot.sendMessage(chat_id=chat_id, text="Error happened %s" % e)
+
+    def _check_services(self, chat_id, username, get_status=False):
+        states = []
+        for file in glob.glob("cockpit/%s/services/**/*.yaml" % username, recursive=True):
+            with open(file, 'r') as f:
+                y = yaml.load(f)
```
zaibon: There is easier way to inspect the status of a service then parsing manually the state file.
A service object has a state object attached to it.
e.g:
```
repo =  j.atyourservice.repos['cockpit']
service = repo.findService('...')
service.state.methods
```
`service.state.methods` returns a dict that contains the status for all methods of the service.
````
{'init': 'OK', 'install' : 'OK'}
```

lib/JumpScale/clients/cockpit/Asker.py
```diff
@@ -269,6 +269,30 @@ def validate(input):
         portal_password = self.askChoice("Please enter a portal password")
         return portal_password
 
+    def ask_organization(self):
+        def validate(input):
+            return True
+        organization = self.askString("Please enter organization name at itsyou.online")
+        return organization
+
+    def ask_client_id(self):
```
zaibon: Why do we ask client id and secret to the user directly ?

lib/JumpScale/clients/cockpit/Asker.py
```diff
@@ -269,6 +269,30 @@ def validate(input):
...
+        def validate(input):
+            return True
+        client_id = self.askString("Please enter the client ID")
+        return client_id
+
+    def ask_client_secret(self):
```
zaibon: same

lib/JumpScale/clients/cockpit/Asker.py
```diff
@@ -269,6 +269,30 @@ def validate(input):
...
+        def validate(input):
+            return True
+        client_secret = self.askString("Please enter the client secret")
+        return client_secret
+
+    def ask_ityou_username(self):
```
zaibon: same

lib/JumpScale/clients/cockpit/CockpitDeployer.py
```diff
@@ -451,3 +471,19 @@ def deploy(self):
         container_cuisine.core.file_write('/root/.ssh/authorized_keys', key_pub)
 
         return cockpit_repo_path
+
+    def oauth(self, chat_id):
```
zaibon: I don't get this. Why do we do a request on the flask server ouself.
flask server is only there as a call back endpoint for it'syou.online. we don't need to query it ourself.

ashraffouda: this for the server to construct the url cuz the server will keep track for this request

lib/JumpScale/clients/cockpit/Asker.py
ashraffouda: according to this script https://gist.github.com/robvanmieghem/d8606a18aec0b495a147d2ce8f0bc533
we need the client secret to get the user token

zaibon: client secret and client_id are private for the bot. these are not information that you get from the user.

rkhamis:
@ashraffouda @zaibon what's the status of this PR?

PR #228, opened 2016-05-19
Remove SALObject 

xmonader:
Getting rid of SALObject.


lib/JumpScale/sal/aoe/AOEManager.py
```diff
@@ -68,7 +68,7 @@ def __repr__(self):
         return str(self)
 
 
-class AOEManager(SALObject):
+class AOEManager():
```
rkhamis: Could you make them class Classname:
Empty brackets are pointless.

lib/JumpScale/sal/aysfs/AysFs.py
```diff
@@ -2,7 +2,7 @@
 from JumpScale import j
 import os
 
-from sal.base.SALObject import SALObject
+
 
 
 class AysFsFactory(object):
```
rkhamis: We no longer need to inherit from Object in python3. The joys of new style classes!

PR #229, opened 2016-05-19
bug fix installing redis

abdulrahmantkhalifa:
when jumpscale uses cuisine to install
redis on start up , it tries to overwrite
the binary already in use to run jumpscale
causing an error , this is avoided by the
--no-clobber option which if exits will not replace


ahussein:
Did you test this fix with the docker builder ?

abdulrahmantkhalifa:
yes, testing now

ahussein:
Let me know how the tests goes please

PR #237, opened 2016-05-25
Host file test

ashraffouda:


tests/sal/HostFile.py
```diff
@@ -0,0 +1,45 @@
+from JumpScale import j
+hostsfactory=j.sal.hostsfile
+from nose.tools import assert_equal, assert_not_equal, assert_raises, raises, assert_in, assert_not_in
+import random
+import socket
+import struct
+
+class TestHostFile():
```
ahussein: why not extending unittest.TestCase ? what kind of testrunner you are using ?

tests/sal/HostFile.py
```diff
@@ -0,0 +1,45 @@
...
+    def test_remove(self):
+        #generate random ip
+        ip = socket.inet_ntoa(struct.pack('>I', random.randint(1, 0xffffffff)))
+        domains = ["remove.test", "remove2.test"]
+        hostsfile = hostsfactory.get()
+        hostsfile.set(ip, domains)
+        assert_equal(hostsfile.existsIP(ip), True)
+        hostsfile.remove(ip)
+        assert_equal(hostsfile.existsIP(ip), False)
```
ahussein: unittests should not affect the system, updating files on the system should not happen. please have a look at how to use mock objects: https://docs.python.org/3/library/unittest.mock.html

ahussein: we also need to test corner cases
remove ip address that doesnt exist..etc
failure scenarios

ahussein: same comments on the rest of the test cases in this file

tests/sal/SystemFS.py
```diff
@@ -0,0 +1,331 @@
+from JumpScale import j
```
ahussein: Same comments as the other file
please remove all the commented code
please remove all the functions that are not used like bomify, stripbom, ...

tests/sal/HostFile.py
xmonader: nosetests. Nose recognizes the test methods starting with "test"

PR #244, opened 2016-05-27
Consumption

zaibon:
improve AYS consumption management


grimpy:
ovc client part looks good ot me

rkhamis:
We should eventually at least make sure that we don't unconsume parents.
Potentially dangerous.

Issue #247, opened 2016-05-28
no more color support on ubuntu 16.04


despiegk:
![image](https://cloud.githubusercontent.com/assets/6021844/15626326/b2753164-24c2-11e6-8647-cc914f36cd1a.png)

PR #250, opened 2016-05-30
bug fix , colors now appear in js on 1510 and 1604

abdulrahmantkhalifa:
https://github.com/Jumpscale/jumpscale_core8/issues/247


xmonader:
Works for me.

PR #254, opened 2016-05-31
make sure to install lsb-release package

ahussein:


ahussein:
will be fixed in https://github.com/Jumpscale/jumpscale_core8/pull/255

PR #260, opened 2016-06-01
fix reinstall js bug in cuisine portal install method 

abdulrahmantkhalifa:
-fix bug in jumpscale_installed to check for js and js8
-fix bug in portal install to check if jumpscale is installed
before reinstalling it


lib/JumpScale/tools/cuisine/apps/CuisinePortal.py
```diff
@@ -25,7 +25,9 @@ def install(self, start=True, mongodbip="127.0.0.1", mongoport=27017, influxip="
         """
         self.cuisine.bash.environSet("LC_ALL", "C.UTF-8")
         if not self.cuisine.core.isMac:
-            self.cuisine.installerdevelop.jumpscale8()
+            if not self.cuisine.installer.jumpscale_installed():
+                import ipdb; ipdb.set_trace()
```
rkhamis: debugger!

PR #266, opened 2016-06-02
Geodns

zaibon:


lib/JumpScale/tools/cuisine/CuisineGeoDns.py
```diff
@@ -163,6 +164,14 @@ def stop(self, name="geodns_main"):
         """
         self.cuisine.processmanager.stop(name)
 
+    @property
+    def domains(self):
+        domains = []
+        for path in self.cuisine.core.fs_find('$cfgDir/geodns/dns/',type='f', pattern='*.json', recursive=False):
```
rkhamis: Not sure of the flow here, but shouldn't you make sure it's installed first? Otherwise, geodns won't be available in $cfgDir yet.

zaibon: Other method of the class doesn't check that.
I can just check if the directory exists and if not return an empty list.

Issue #246, opened 2016-05-28
think actions broken in cuisine


despiegk:
lots of methods are executed over & over again
the actions optimization avoiding to do things double is broken

mie00:
Probably they are executed again because the action decorator has a kwarg `force=True` which forces the actions to be reexecuted when called again.

despiegk:
so I guess we need to put it on false & test if it works fine
today its way too slow
of make sure the required actions are called with force=False

mie00:
Manually testing it actions when `force=False`, it seems to be working fine

However, these methods are the methods that are defined with `force=True`, and for them that seems logical
```
/cuisine/CuisineCore.py:
sudo, run, run_script, tmux_execute_jumpscript, execute_jumpscript, execute_python

/cuisine/CuisineSSH.py:
test_login, test_login_pushkey, keygen, authorize, unauthorize, unauthorizeAll, sshagent_add, sshagent_remove

/cuisine/apps/CuisinePortal.py:
start, stop

/avahi/Avahi.py:
reload, removeService, getServices
```

Also there were missing decorators on some cuisine application builders which should be fixed by #272, but I think they are not directly related to this issue.

So if the problem still exists, please point out how it can be reproduced.

PR #272, opened 2016-06-06
add action decorator for some apps

mie00:
See #246


lib/JumpScale/tools/cuisine/apps/CuisineInfluxdb.py
```diff
@@ -23,6 +23,7 @@ def __init__(self, executor, cuisine):
         self.executor = executor
         self.cuisine = cuisine
 
+    @actionrun(action=True)
```
rkhamis: I'm okay with this, makes sense
but if start=true, the expected behavior is changed, because services won't start.

lib/JumpScale/tools/cuisine/apps/CuisinePortal.py
```diff
@@ -17,6 +17,7 @@ def __init__(self, executor, cuisine):
         self.portal_dir = self.cuisine.core.args_replace('$appDir/portals/')
         self.main_portal_dir = j.sal.fs.joinPaths(self.portal_dir, 'main')
 
+    @actionrun(action=True)
```
rkhamis: same for start

lib/JumpScale/tools/cuisine/apps/CuisineInfluxdb.py
mie00: Maybe we can split it into two methods, an internal method that only installs the service and an external method that calls the internal one and then starts the service, and like that we can only add the decorator to only the internal method.

rkhamis: Nice solution

PR #275, opened 2016-06-06
check if path is none first

xmonader:


lib/JumpScale/sal/fs/SystemFS.py
```diff
@@ -443,13 +443,14 @@ def removeDirTree(self, path, onlyLogWarningOnRemoveError=False):
         """Recursively delete a directory tree.
             @param path: the path to be removed
         """
+        if path is None:
+            raise ValueError('Path is None in system.fs.removeDir')
```
rkhamis: please use `j.exceptions`

PR #291, opened 2016-06-09
fix typo and remove pudb

xmonader:


rkhamis:
instead of removing the pudb, could you raise an error?
You don't wan to end there silently

PR #294, opened 2016-06-09
fixing extensions to include a dot

xmonader:


rkhamis:
getFileExtension already back to its old behavior. This fix is no longer needed

PR #303, opened 2016-06-13
GitHub tools refactor

muhamadazmy:


lib/JumpScale/clients/github/Issue.py
```diff
@@ -39,87 +30,85 @@ def ddict(self):
...
             # no dict yet, fetch from github
             self.load()
+        # we lazy load the comments. so it's only loaded when accesses
+        self._ddict['comments'] = self.comments
         return self._ddict
 
     @property
     def comments(self):
-        return self.ddict["comments"]
+        with self._lock:
```
FastGeert: You can improve this by checking if self._comments is None before taking the lock. But you need to do it again inside the lock.

PR #323, opened 2016-06-20
bug fix in cuisine redis

abdulrahmantkhalifa:
redis client has no attribute called self.cuisine its called self._cuisine


lib/JumpScale/tools/cuisine/apps/CuisineRedis.py
```diff
@@ -71,8 +71,8 @@ def build(self,name="main",ip="localhost", port=6379, maxram=200, appendonly=Tru
 
         self.cuisine.bash.addPath(j.sal.fs.joinPaths(self.cuisine.core.dir_paths["base"], "bin"), action=True)
 
-        j.clients.redis.executor=self.executor
-        j.clients.redis.cuisine=self.cuisine
+        j.clients.redis._executor=self.executor
```
rkhamis: get an instance of the client, and in the init, set these values.
this code will break and is very hackish.

also, it won't fix your issue because the redis client still sets it to local cuisine. You need to implement a proper constructor

lib/JumpScale/clients/redis/Redis.py
```diff
@@ -70,6 +70,10 @@ def __init__(self):
         self._config = {}
         self._cuisine = j.tools.cuisine.get()
 
+    def getInstance(self, cuisine):
```
rkhamis: please make it "get"

PR #329, opened 2016-06-20
Org issue 286

abdulrahmantkhalifa:
https://github.com/gig-projects/org_development/issues/286


lib/JumpScale/tools/cuisine/CuisineInstallerDevelop.py
```diff
@@ -56,6 +58,9 @@ def python(self):
     def pip(self):
         self.cuisine.installer.base()
         self.python()
+        if self.cuisine.core.isMac:
```
zaibon: so pip support is completely dropped for mac ?

lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1037,7 +1046,12 @@ def run(self,cmd,die=True,debug=None,checkok=False,showout=True,profile=False,re
             cmd = 'echo %s | sudo -S bash -c "%s"' % (passwd, cmd)
         else:
             cmd = 'bash -c "%s"' % cmd
-        rc,out=self.executor.execute(cmd,checkok=checkok, die=False, combinestdr=True,showout=showout)
+
+        path = self.executor.execute("echo $PATH", showout=False)[1]
```
zaibon: I think we can use the cuisine.bash module to read PATH value

PR #334, opened 2016-06-21
add default value to ensure_domain + fix serial logic

rkhamis:


lib/JumpScale/tools/cuisine/CuisineGeoDns.py
```diff
@@ -111,7 +111,7 @@ def save(self):
             self.content["data"][key]["cname"] = val
 
         self.content["ttl"] = self.ttl
-        self.content["serial"] = self.serial
+        self.content["serial"] = int(self.serial) + 1
```
rkhamis: `serial` could possibly be None. Maybe change the method definition to have `serial=0` instead of None?

PR #340, opened 2016-06-22
Ays 246

ahussein:


lib/JumpScale/baselib/atyourservice/AtYourServiceFactory.py
```diff
@@ -66,11 +66,12 @@ def reset(self):
         self._repos = {}
         j.dirs._ays = None
 
+
     @property
     def repos(self):
         if self._repos == {}:
             for path in j.atyourservice.findAYSRepos():
-                name = j.sal.fs.getBaseName(path)
+                name = path.strip('/').replace('/', '.')
```
rkhamis: but this means that accessing the repo will later become very difficult. 
will it be `full.path.reponame` ?

ahussein: Yes, it will be opt.code.test1.test for example, we can git rid of the opt.code part and assume that everything will be under /opt/code ...Having a full name with some kind of namespace is a clean way of defining the name of the repo. normally you would list the repos before using one of them so I do not think it will be difficult, but other suggestions are welcomed ofc!

rkhamis: This will break a lot of what we already have. I agree with the method, but let's see if there's a more user-friendly way of doing this. 

How about if the repos are indexed with the path but gotten with the basename. we can allow getting with path as well.
If there's a conflict in the basename, it would get the first available one or throw an error.

So it would look something like:
```
repos = {
     '..../test1/test/': repoobject1,
     '..../test2/test/: repoobject2,
}

j.atyourservice.get(test) ---> gets first match or throws error that too many are found
j.atyourservice.get(path='..../test2/test') ---> gets correct repo
```

Not loving this approach either. Can be unpredictable and can cause issues. What do you think?

rkhamis: Let's raise an error when trying to get a repo with name when multiple repos with the same name exist.
Error should be helpful. Something like "multiple AYS repos with name "%s" are found under locations " " and " ". Please use j.atyourservice.get(path=<path>) instead"

ahussein:
updated the PR

lib/JumpScale/baselib/atyourservice/AtYourServiceFactory.py
```diff
@@ -54,24 +54,49 @@ def getTester(self, name="fake_IT_env"):
         return AtYourServiceTester(name)
 
     def get(self, name, path=""):
+        """
```
rkhamis: shouldn't name be an optional argument in this case?

ahussein:
Updated the PR

rkhamis:
@ahussein these imports broke AYS. Please test using this AYS repo: https://github.com/Jumpscale/jumpscale_ays8_testenv/tree/master/fake_IT_env

I reverted the merge
Reopen PR when ready.

PR #368, opened 2016-06-28
mac: bug fix in CuisineInfluxdb

abdulrahmantkhalifa:
the config file is not copied and so is not available when trying to use
to create the process exec


rkhamis:
Reverted because it breaks AYS

rkhamis:
Re-applied. wrong PR

PR #369, opened 2016-06-28
fix combinestdr for the sshclient

### LINKED ISSUE #348: check that sshclient/cuisine sends stderror back to client
I think when there is something on stderror it doesn't get returned to client
lets check

xmonader:
Fixes #348


lib/JumpScale/clients/ssh/SSHClient.py
```diff
@@ -298,6 +298,9 @@ def run(self):
         if retcode == 0 and out == '' and len(err) > 0:
             out = err
 
+        if combinestdr == True:
+            if err != '':
+                out = "%s\nSTDERR:%s"%(out, err)
```
rkhamis: `if combinestdr and err.strip():` is nicer, no?

zaibon: This is not the expected behaviour. combine implies that the output of the command should include both stdout and stderr in the same buffer no ?
This output is not easily parsable.

Or we merge stdout and stderr and return that or we return stdout and stderr in two variables.

rkhamis: We discussed that a lot, but that would mean that the same method would have different outputs for different parameters.

zaibon: exactly and I don't like that either. Doing so would imply a big refactor off all the code that uses the ssh client, do we want to go that path ?

But this code just destroy the output of the sshclient, and I'm sure code that use combined output don't expect this format.

Issue #374, opened 2016-06-28
cuisine coreos does not install 


abdulrahmantkhalifa:
## issue 
the config file names used  in cuisine.apps.core.build , have been deprecatred , (the file names changed )

## solution 
use the new names

PR #375, opened 2016-06-28
used new core config names

abdulrahmantkhalifa:


abdulrahmantkhalifa:
https://github.com/Jumpscale/jumpscale_core8/issues/374

PR #380, opened 2016-06-29
 bug fix + typo fix in cuisine golang

abdulrahmantkhalifa:


lib/JumpScale/tools/cuisine/CuisineGolang.py
```diff
@@ -60,13 +60,17 @@ def install(self):
 
     @actionrun(action=True)
     def goraml(self):
+        import ipdb;ipdb.set_trace()
```
rkhamis: DEBUGGER!

PR #402, opened 2016-07-11
support for cygwing

abdulrahmantkhalifa:
install support for windows, installation finshes until js , installjsdepps does not work as cuisine doesnt not load properly


abdulrahmantkhalifa:
https://github.com/gig-projects/org_development/issues/287

PR #413, opened 2016-07-18
Out err fix

khaledkbadr:


zaibon:
How is this will ever work with SSH Executor ? SSHClient has not been updated to return the stdout and stderr in different variable

PR #421, opened 2016-07-19
Execute fix

khaledkbadr:


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1098,7 +1098,7 @@ def run(self, cmd, die=True, debug=None, checkok=False, showout=True, profile=Fa
         out = out.strip()
 
         if check_is_ok:
-            return out.find("**OK**") != -1
+            return rc, out.find("**OK**") != -1, err
```
rkhamis: Are we still appending `**OK**` to the result?

khaledkbadr: yup, didn't touch it. I don't know it's effect and it seems to be not straight. I prefer to fix it in another iteration

PR #422, opened 2016-07-19
Check is ok refactor to its own function to remove redundancy

khaledkbadr:


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -1098,7 +1096,7 @@ def run(self, cmd, die=True, debug=None, checkok=False, showout=True, profile=Fa
         out = out.strip()
 
         if check_is_ok:
-            return out.find("**OK**") != -1
+            return rc, out.find("**OK**") != -1, err
```
rkhamis: I'd rather we return `rc, not rc, error` given that you've already removed a lot of the `echo **OK**`s

khaledkbadr:
Merge #421 before merging this pr

PR #435, opened 2016-07-24
bug fix in error condition object

abdulrahmantkhalifa:


lib/JumpScale/core/errorhandling/ErrorConditionHandler.py
```diff
@@ -218,7 +218,7 @@ def parsePythonExceptionObject(self,exceptionObject,tb=None):
             eco.tracebackSet(tb,exceptionObject)
 
             if len(eco.traceback)>10000:
-                eco.traceback=eco.traceback[:10000]
+                eco._limit = 10000
```
mie00: Can't we get rid of this if statement at all?

lib/JumpScale/core/errorhandling/ErrorConditionObject.py
```diff
@@ -86,12 +87,15 @@ def tracebackSet(self,tb,exceptionObject):
         # tb=e.__traceback__
         type=None        
         tblist=traceback.format_exception(type, exceptionObject, tb)
-        
         self._traceback=""
 
         ignore=["click/core.py","ipython"]
 
-        for item in tblist:
+        tblist.reverse()
```
mie00: like that it will be printed in the reverse order, right?

PR #441, opened 2016-07-25
Make sure the sshclient gets invoked so session is active

rkhamis:
Needs more testing


lib/JumpScale/clients/ssh/SSHClient.py
```diff
@@ -25,10 +25,15 @@ def reset(self):
     def get(self, addr, port=22, login="root", passwd=None, stdout=True, forward_agent=True, allow_agent=True, look_for_keys=True,
             timeout=5, key_filename=None, passphrase=None, die=True, usecache=True):
         key = "%s_%s_%s_%s" % (addr, port, login, j.data.hash.md5_string(str(passwd)))
+        if key in self.cache and usecache:
+            try:
+                if not self.cache[key].transport.is_active():
+                    usecache = False
```
grimpy: if its not active anymore shouldnt we just pop it out of the cache?

grimpy: nevermind previous comment just noticed that even when setting usecache to false you still store the new connectin in the cache object.

rkhamis: yes. yes, we can.

zaibon: Can we document the argument so people understands what the behaviour is

lib/JumpScale/tools/executor/ExecutorSSH.py
```diff
@@ -73,7 +74,12 @@ def sshclient(self):
                 else:
                     raise j.exceptions.RuntimeError("Could not find key:%s"%path)
 
-                self._sshclient.ssh_authorize("root",pubkey)
+            self._sshclient = j.clients.ssh.get(self.addr, self.port, login=self.login, passwd=self.passwd,
+                                                allow_agent=self.allow_agent, look_for_keys=self.look_for_keys,
+                                                key_filename=path, passphrase=None,
+                                                timeout=self.timeout, usecache=False)  # TODO: add passphrase
```
despiegk: please check @todo in this code see on branch ays_capnp or master branch
we need to document pushkey pubkey behaviour
is all very confusing

this should have been separate method for push key

the pubkey why do we need this?

also the =None ="" ... behaviour is wrong DO NOT USE
if pubkey:
   ....

needs to be all specific

PR #447, opened 2016-07-27
Minor fix to issues_by_priority and issues_by_state

khaledkbadr:


khaledkbadr:
I've noticed that if the an issue has no label we assign `priority_normal` label to it, then check if it's in the repo labels which raises error if it's not found. So basically we force the repo to have `priority_normal` as a label. 

Is it supposed to work like that?

PR #457, opened 2016-08-02
Connectviaproxy

xmonader:


despiegk:
why do we need this?

xmonader:
Because sometimes we need to ssh to a machine that doesn't have a public IP via proxy.

despiegk:
we will not use this, code is bad, not documented & very badly structured
its all very confusing

PR #459, opened 2016-08-02
added installjsdep method back to install.sh

despiegk:
wrong fix, CANNOT do 
js 'j.tools.cuisine.local.installerdevelop.installJS8Deps()'


install/install.sh
```diff
@@ -132,3 +135,4 @@ curl -k https://raw.githubusercontent.com/Jumpscale/jumpscale_core8/$branch/inst
 
 cd $TMPDIR
 python3 bootstrap.py
+js 'j.tools.cuisine.local.installerdevelop.installJS8Deps()'
```
despiegk: this is VERY wrong !!!!!!!!!!
please contact Azmy to know why

PR #466, opened 2016-08-04
Fix unittest

khaledkbadr:


ahussein:
the unittest branch was not up to date with the latest master changes and this makes this diff too huge and contains too many unrelated changes. I will close this PR and create new one after updating the unittest branch with the latest changes from master.

PR #470, opened 2016-08-05
Use path and not service key name during getService

maxux:


maxux:
This fix ays install error related to:
```
STEP:1, ACTION:install@action()
def install(self, service):
 [...]
```

PR #471, opened 2016-08-07
remove sslVerify=False

mie00:


zaibon:
Why do we want to remove this ?

mie00:
Because of the possibility for MITM attacks against jumpscale.

PR #474, opened 2016-08-07
use findAYSRepos instead of repos

khaledkbadr:


lib/JumpScale/core/main/Application.py
```diff
@@ -308,7 +308,10 @@ def getAppHRDInstanceNames(self,name,domain="jumpscale"):
         """
         returns hrd instance names for specific appname (default domain=jumpscale)
         """
-        names = [service.instance for aysrepo in list(j.atyourservice.repos.values()) for service in list(aysrepo.services.values()) if service.templatename == name]
+        repos = []
+        for path in j.atyourservice.findAYSRepos('/opt/'):
```
zaibon: I would use `j.dirs.codeDir` instead of `/opt/`

PR #515, opened 2016-08-28
Agent7to8

muhamadazmy:
Porting jsagent7 to JS8


apps/jsagent/cmds/AgentCmds.py
```diff
@@ -0,0 +1,157 @@
...
+            try:
+                running = redisinstance.pint()
+            except:
+                running = False
+            if not running:
+                return False
+        return True
+
+    def checkRedisSize(self, session=None):
+        redisinfo = j.clients.redisworker.redis.info().split('\r\n')
```
rkhamis: Should this not be `j.legacy.redisworker`?

apps/jsagent/cmds/JumpscriptsCmds.py
```diff
@@ -0,0 +1,152 @@
...
+# from rq import Queue
+# import JumpScale.baselib.redisworker
+from JumpScale.legacy.redisworker.RedisWorker import RedisWorkerFactory
+import crontab
+from JumpScale.legacy.jumpscripts.JumpscriptFactory import Jumpscript
+
+
+class JumpscriptsCmds():
+    def __init__(self, daemon=None):
+        self.log = j.logger.get('JumpscriptsCmds')
```
rkhamis: Inconsequential, but I'd prefer to keep logging consistently under `j.legacy.agent.jumpscriptsCmds` for instance

apps/jsagent/jsagent.py
```diff
@@ -0,0 +1,232 @@
...
+                            self.log.info("Process %s has stopped" % p)
+                            p.kill()
+                            self.processes.remove(p)
+
+            time.sleep(1)
+            if len(self.processes) == 0:
+                self.log.info("no more children")
+                # return
+
+parser = cmdutils.ArgumentParser()
```
rkhamis: We use `click` http://click.pocoo.org/5/commands/ in JS8

apps/jsagent/cmds/AgentCmds.py
muhamadazmy: Yip, you are right

apps/jsagent/cmds/JumpscriptsCmds.py
muhamadazmy: I don't know exactly how the j.logger.get work, if it should take full name or just the class name. If it should take the full extension name like `j.legacy.agent.jumpscriptsCmds` then you are totally right.

apps/jsagent/jsagent.py
muhamadazmy: It's ported from js7 and I assume this app should not live inside js8 for long time. So that's is an un-needed effort imho.

PR #570, opened 2016-11-22
added the cuisine ays_bot to install and configure the ays bot

abdulrahmantkhalifa:
https://github.com/Jumpscale/jscockpit/issues/245

### DIFF
--- a/lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
+++ b/lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
@@ -1,0 +1,111 @@
+from JumpScale import j
+
+
+app = j.tools.cuisine._getBaseAppClass()
+
+
+class CuisineAysBot(app):
+    NAME = ''
+
+    def isInstalled(self):
+        return self._cuisine.core.file_exists(
+            "$appDir/ays_bot/telegram-bot") and self._cuisine.core.file_exists('$cfgDir/ays_bot/config.toml')
+
+    def __init__(self, executor, cuisine):
+        self._cuisine = cuisine
+        self.executor = executor
+        self._configured = False
+
+    def install(self, start=True, link=True, reset=False):
+        """
+        please configure first
+        Install aysbot
+        If start is True, token g8_addresses, dns and oauth should be specified
+        """
+        if not reset and self.isInstalled():
+            return
+
+        self._cuisine.bash.environSet("LC_ALL", "C.UTF-8")
+        if not self._cuisine.core.isMac and not self._cuisine.core.isCygwin:
+            self._cuisine.development.js8.install()
+            self._cuisine.development.pip.packageUpgrade("pip")
+
+        self.install_deps()
+        self._cuisine.development.git.pullRepo('https://github.com/Jumpscale/jscockpit.git')
+        if link:
+            self.link_code()
+        else:
+            self.copy_code()
+
+        if start:
+            self.start()
+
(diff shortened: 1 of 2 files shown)


lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, executor, cuisine):
+        self._cuisine = cuisine
+        self.executor = executor
+        self._configured = False
+
+    def install(self, start=True, link=True, reset=False):
+        """
+        please configure first
+        Install aysbot
+        If start is True, token g8_addresses, dns and oauth should be specified
```
zaibon: token g8_addresses, dns from the comment doesn't exist in the method signature

lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        if not reset and self.isInstalled():
+            return
+
+        self._cuisine.bash.environSet("LC_ALL", "C.UTF-8")
+        if not self._cuisine.core.isMac and not self._cuisine.core.isCygwin:
+            self._cuisine.development.js8.install()
```
zaibon: we don't install dependencies from the cuisine module itself.
user should make sure it install the dependencies first.

lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
```diff
@@ -0,0 +1,111 @@
...
+        token: telegram bot token received from @botfather
+        oauth: dict containing
+               host
+               oauth
+               client_id
+               client_secret
+               itsyouonline.host
+        cfg: full dict with config optional
+        """
+        import ipdb; ipdb.set_trace()
```
zaibon: remaining debug

lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
```diff
@@ -0,0 +1,111 @@
...
+                       'organization': client_id,
+                       'client_id': client_id,
+                       'client_secret': client_secret,
+                       'redirect_uri': redirect,
+                       'itsyouonlinehost': itsyouonlinehost
+                       }
+                   }
+
+        config = j.data.serializer.toml.dumps(cfg)
+        self._cuisine.core.dir_ensure("$cfgDir/ays_bot")
```
zaibon: the path to the config should be `$cfgDir/ays_bot/{instance_name}` otherwise we can't install two ays_bot on the same system. need to update in the code above too.

lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
zaibon: This is still not good, this line should be removed completely. We don't want to install jumscale automatically from here.

lib/JumpScale/tools/cuisine/apps/CuisineAysBot.py
zaibon: I would prefer we have an extra level of directory instead of naming the config file differently. so
`$cfgDir/ays_bot/{instance_name}/config.toml`
instead of 
`$cfgDir/ays_bot/{instance_name}.toml`

zaibon:
merged into js8.1

PR #571, opened 2016-11-22
Adds method to check for blueprints is a valid yaml format and report…

ashraffouda:
… errors


zaibon:
merged into js8.1

PR #610, opened 2016-12-04
Ays noexec

zaibon:
if no exec mode is enable, the actions are not really executed.
Everything stil works notrmaly but AYS doesn't impact on reallity

### DIFF
--- a/lib/JumpScale/baselib/jobcontroller/Run.py
+++ b/lib/JumpScale/baselib/jobcontroller/Run.py
@@ -55,15 +55,30 @@ class RunStep:
             res.append(job_model.objectGet())
         return res
 
+    def _fake_exec(self, job):
+        job.model.dbobj.state = 'ok'
+        action_name = job.model.dbobj.actionName
+        # if the action is a reccuring action, save last execution time in model
+        if action_name in job.service.model.actionsRecurring:
+            job.service.model.actionsRecurring[action_name].lastRun = j.data.time.epoch
+
+        service_action_obj = job.service.model.actions[action_name]
+        service_action_obj.state = 'ok'
+        job.save()
+
     def execute(self):
         processes = {}
         for job in self.jobs:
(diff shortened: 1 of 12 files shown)


lib/JumpScale/baselib/jobcontroller/Run.py
```diff
@@ -55,15 +55,30 @@ def jobs(self):
...
 
+    def _fake_exec(self, job):
+        job.model.dbobj.state = 'ok'
+        action_name = job.model.dbobj.actionName
+        # if the action is a reccuring action, save last execution time in model
+        if action_name in job.service.model.actionsRecurring:
+            job.service.model.actionsRecurring[action_name].lastRun = j.data.time.epoch
+
+        service_action_obj = job.service.model.actions[action_name]
+        service_action_obj.state = 'ok'
```
zaibon: I though that would be still use full to update the state like if the action succeed. So at least the creation of the run would still behave the same as real execution.

rkhamis: There's nothing currently constraining that no-exec cannot be set/unset within a repo.
This will create some problems like:

- setting `no-exec` after doing some runs can result in false `ok` states
- unsetting `no-exec` after some runs will make AYS skip "falsely ok" services.

Suggestion is to set `no-exec` on repo model and no mixing of `exec` and `no-exec` be allowed.

rkhamis:
There's nothing currently constraining that no-exec cannot be set/unset within a repo.
This will create some problems like:

- setting no-exec after doing some runs can result in false ok states
- unsetting no-exec after some runs will make AYS skip "falsely ok" services.

Suggestion is to set no-exec on repo model and no mixing of exec and no-exec be allowed.

zaibon:
Updated my PR.

Now we can enable of disable the noexec mode on a repo level.

PR #612, opened 2016-12-05
Refactor Flist tools to be usable by core0 0.10.0

zaibon:
also add a small main method to the flist module to easily create a flist


rkhamis:
Should this be in 8.1.0?
I can already hear people screaming about new imports

zaibon:
No indeed this doesn't need to be in 8.1.0, you can report it for later.
I need a review from @maxux too

despiegk:
same comment here, plist is being implemented differently, lets not loose too much time here

muhamadazmy:
I think we need to have some sort of filtering on what files need to be uploaded and added in the flist. For examples, `.h` files, `man` and `doc` pages. `.pyc` files `icons`, `wallpapers`, `images`, and anything that won't be needed in a production environment.

This will reduce the flist size, and also the number of files uploaded to the store.

rkhamis:
@maxux @zaibon I changed the base to be 8.2.0. Can you update this branch or delete it?

maxux:
Need @zaibon confirmation, but this can be probably dropped. The 8.2.0 will use the new capnp flist format and lot of commits here are not relevant anymore.

PR #618, opened 2016-12-06
Make sure values defined in schema.hrd are not lowered

rkhamis:
Default values were always lowered, which is very troublesome

### DIFF
--- a/lib/JumpScale/data/hrd/HRDSchema.py
+++ b/lib/JumpScale/data/hrd/HRDSchema.py
@@ -146,7 +146,7 @@ class HRDSchema:
             name = name.strip()
 
             tags = j.data.tags.getObject(tagstr)
-            tags_dict = {key.lower():value for key, value in tags.getDict().items()}
+            tags_dict = {key.casefold():value for key, value in tags.getDict().items()}
 
             if "type" in tags_dict:
                 ttype = tags_dict.get("type").strip().lower()
--- a/lib/JumpScale/data/hrd/HRDSchema.py
+++ b/lib/JumpScale/data/hrd/HRDSchema.py
@@ -144,19 +144,19 @@ class HRDSchema:
 
             name, tagstr = line.split("=", 1)
             name = name.strip()
-            tagstr = tagstr.lower()
 
             tags = j.data.tags.getObject(tagstr)
+            tags_dict = {key.lower():value for key, value in tags.getDict().items()}
 
-            if tags.tagExists("type"):
-                ttype = tags.tagGet("type").strip().lower()
+            if "type" in tags_dict:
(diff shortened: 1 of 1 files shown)


lib/JumpScale/data/hrd/HRDSchema.py
```diff
@@ -177,16 +177,16 @@ def process(self, content):
 
             hrdtype.typeclass = j.data.types.getTypeClass(ttype)
 
-            if tags.labelExists("list"):
+            if "list" in tags_dict:
```
grimpy: labels are not in this dict

lib/JumpScale/data/hrd/HRDSchema.py
```diff
@@ -215,38 +215,38 @@ def process(self, content):
...
-                c = tags.tagGet("parent")
+            if "parent" in tags_dict:
+                c = tags_dict.get("parent")
                 hrdtype.consume_nr_min = 1
                 hrdtype.consume_nr_max = 1
                 hrdtype.consume_link = c
                 hrdtype.parent = c
 
-            if tags.labelExists("auto"):
+            if "auto" in tags_dict:
```
grimpy: label not a tag

lib/JumpScale/data/hrd/HRDSchema.py
```diff
@@ -215,38 +215,38 @@ def process(self, content):
...
                 hrdtype.auto = True
 
-            if tags.labelExists("optional"):
+            if "optional" in tags_dict:
```
grimpy: label not a tag

muhamadazmy:
Won't it easier (and probably cleaner) that the tagExists method (or labelExsts) do a key.lower() itself. so you don't have to copy all the tags to a dict and then check against the dict ?

lib/JumpScale/data/hrd/HRDSchema.py
rkhamis: In the dict, I see
```
tags_dict
{'auto': True, 'type': 'str'}
```
So it's in there.

rkhamis:
@muhamadazmy Indeed. But that thing is used allover and I don't want to change its behavior

grimpy:
what would be better is that you use propercasing everywhere

rkhamis:
@grimpy absolutely. Now ask me if I want to break all existing actors and blueprints.

muhamadazmy:
i think the tag key should be case insensitive, so it's okay if we apply it everywhere. but u are the owner of the product, so u decide.

Also i think u should use `str.casefold()` instead of `str.lower()` for proper case-insinsitive comparison.

rkhamis:
@muhamadazmy first time I know about casefold. Very cool. Done!

PR #624, opened 2016-12-07
Add flist archiver argument forwarder and improve hashing

maxux:


despiegk:
there will be a new flist implementation based on a key value stor, don't put too much time in this one

PR #626, opened 2016-12-07
fixes for hanging processes #293

xmonader:
Using core run to getPIDs instead of j.sal.process.execute


lib/JumpScale/sal/process/SystemProcess.py
```diff
@@ -1583,7 +1583,8 @@ def getPidsByFilterSortable(self, filterstr, sortkey=None):
             cmd = "ps aux --sort={sortkey} | grep '{filterstr}'".format(filterstr=filterstr, sortkey=sortkey)
         else:
             cmd = "ps ax | grep '{filterstr}'".format(filterstr=filterstr)
-        rcode, out = j.sal.process.execute(cmd)
+        lc = j.tools.cuisine.local
```
zaibon: Using cuisine from a sal is a bad idea. SAL should be the lower layer and cuisine should use the SALs, not the other way round.

Also this is done to fix a bug in the portal. I don't see why we change jumpscale core to fix a portal issue like this one.
Portal is a gevent monkey patch process, you can't exepct code not written for that to work. So https://github.com/Jumpscale/jscockpit/issues/293 should be solved differently

PR #628, opened 2016-12-07
Autorecover logging cfg

rkhamis:


install/InstallTools.py
```diff
@@ -199,9 +197,11 @@ def writeEnv(self):
                             os.environ[item] += "/"
```
ashraffouda: why don't we have a default values for config in case the env variables does not exist?

PR #629, opened 2016-12-07
OpenvCloud client: make it possible to refresh jwt tokens

grimpy:


lib/JumpScale/clients/openvcloud/Client.py
```diff
@@ -3,10 +3,24 @@
...
 import os
+import requests
 
 CACHETIME = 60
 
 
+def refresh_jwt(jwt, payload):
+    if payload['iss'] == 'itsyouonline':
+        refreshurl = "https://itsyou.online/v1/oauth/jwt/refresh"
+        response = requests.get(refreshurl, headers={'Authorization': 'beater {}'.format(jwt)})
```
zaibon: typo: should be `bearer` not `beater`

lib/JumpScale/clients/oauth/OauthInstance.py
```diff
@@ -42,3 +54,79 @@ def url(self):
...
         return '%s?%s' % (self.addr, urllib.parse.urlencode(params))
+
+    def getAccessToken(self, code, state):
+        payload = {'code': code, 'client_id': self.id, 'client_secret': self.secret,
+                   'redirect_uri': self.redirect_url, 'grant_type': 'authorization_code',
+                   'state': state}
+        result = requests.post(self.accesstokenaddress, data=payload, headers={
+            'Accept': 'application/json'})
+
+        if not result.ok or 'error' in result.json():
```
zaibon: IYO doens't return valid json during an error, so the call to json() will crash

PR #642, opened 2016-12-14
Adds getBranch method

ashraffouda:
Signed-off-by: Ashraf Fouda <foudaa@greenitglobe.com>


lib/JumpScale/clients/git/GitClient.py
```diff
@@ -84,6 +84,9 @@ def getBranchOrTag(self):
         except:
             return 'branch', self.branchName
 
+    def getBranch(self):
+        return 'branch', self.repo.head.ref.name
+
```
rkhamis: Does this also get latest tag?

lib/JumpScale/clients/git/GitClient.py
```diff
@@ -331,3 +331,12 @@ def patchGitignore(self):
             out = '\n'.join(linesout)
             if out.strip() != inn.strip():
                 j.sal.fs.writeFile(ignorefilepath, out)
+
+    def describe(self):
+        """
+        this method get latest tag or branch
+        """
+        try:
+            return 'tag', j.tools.cuisine.local.core.run('git describe --tags')[1]
```
rkhamis: it makes more sense that a client uses sal instead of cuisine
Consider `j.sal.process.execute`

Issue #644, opened 2016-12-14
get account API shouldn't create account by default


islamTaha12:
Calling the following method from any AYS service will create an account however its a get account.
Example: node ovc service create accounts if it doesn't in cache yet, So this will try to create an account while the enviroment has the account.
https://github.com/Jumpscale/jumpscale_core8/blob/8.1.0/lib/JumpScale/clients/openvcloud/Client.py#L148

**Scenario**
- Create an account by post API /restmachine/cloudbroker/account/create
- Consume this account inside node.ovc service
- Run this blueprint using cockpit

**Actual result**
![screenshot from 2016-12-14 14-42-29](https://cloud.githubusercontent.com/assets/17981089/21182432/ca617792-c20b-11e6-9ef7-71a84382cdd8.png)
  

**Expected result**
node.ove shouldn't create an account.

PR #645, opened 2016-12-14
Removing caching ovc client to always be in sync with env.

rkhamis:
Relates to #644


lib/JumpScale/clients/openvcloud/Client.py
```diff
@@ -218,20 +195,13 @@ def __init__(self, client, model):
...
     def spaces(self):
-        if not self._spaces_cache:
-            # load from api
-            for item in self.client.api.cloudapi.cloudspaces.list():
-                if item['accountId'] == self.model['id']:
-                    self._spaces_cache.set(item, id=item['id'])
-        spaces = []
-        for space in self._spaces_cache:
-            spaces.append(Space(self, space.struct))
+        ovc_spaces = self.client.api.cloudapi.cloudspaces.list()
```
zaibon: why don't we check the on `if item['accountId'] == self.model['id']:` anymore ?

PR #673, opened 2017-01-02
order matters for fixlocals and resets

xmonader:


lib/JumpScale/tools/bash/Bash.py
```diff
@@ -135,9 +135,10 @@ def environ(self):
                     val = val.strip().strip("'").strip("\"")
                     res[name] = val
             self._environ = res
-        self.fixlocale()
         merge = self._environ.copy()
+        self.fixlocale()
```
ashraffouda: I think u should go for fixing `fixlocale()` instead

PR #680, opened 2017-01-04
caddy cuisine

zaibon:
install using the builddir
add BINDIR to default path
fix call to cuisine ufw


lib/JumpScale/tools/cuisine/apps/CuisineCaddy.py
```diff
@@ -16,14 +21,17 @@ def install(self, ssl=False, start=True, dns=None, reset=False):
         @param dns str: default address to run caddy on.
         @param reset bool:  if True this will install even if the service is already installed.
         """
-        if reset is False and self.isInstalled():
+        import ipdb; ipdb.set_trace()
```
rkhamis: debugger!

PR #685, opened 2017-01-05
update cuisine module for tidb

zaibon:
This requires https://github.com/ngaut/rust-rocksdb/pull/72 to be merged in to work properly.


lib/JumpScale/tools/cuisine/CuisineCore.py
```diff
@@ -343,8 +343,7 @@ def file_get_tmp_path(self, basepath=""):
         return x
 
     def file_download(self, url, to="", overwrite=True, retry=3, timeout=0, login="",
-                      passwd="", minspeed=0, multithread=False, expand=False, minsizekb=30,
-                      removeTopDir=False):
+                      passwd="", minspeed=0, multithread=False, expand=False, minsizekb=0, removeTopDir=False):
```
despiegk: @zaibon minsize should be +30 because otherwise can fail on a download which downloads a page not found

PR #688, opened 2017-01-08
initial ayswalkthrough

alichaddad:


rkhamis:
This is a great initial doc.

But maybe it could be a little bit more verbose? Explain what happens internally at each point. Explain why you do each of the steps...

Also, it's usually a lot easier to follow short to-the-point walkthroughs that address a specific point.
For instance:

- repo walkthrough
- blueprint walkthrough
- actor walkthrough
....

PR #702, opened 2017-01-10
8.2.0 ays

despiegk:


lib/JumpScale/baselib/atyourservice81/AtYourServiceDaemon.py
```diff
@@ -149,26 +149,31 @@ def _progagate_event(self, request):
 
         event_type = request['event']
         args = request.get('args', {})
-
+        handler_name = "on_" + event_type
```
zaibon: This completely change the way event are handled.
Here we expect people to implement a method with a specific name. Which mean they can only have 1 method by event.

Before the mapping between method and event is done in the `service.hrd` and gives more flexibility since you can have more then one method for the same event and you can call it whatever you want.

lib/JumpScale/baselib/atyourservice81/AtYourServiceDaemon.py
```diff
@@ -149,26 +149,31 @@ def _progagate_event(self, request):
...
-                    event_obj.lastRun = j.data.time.epoch
-                    service.save()
-
-                    self._workers.apply_async(run_action, (
-                        service.aysrepo.path,
-                        service.model.key,
-                        event_obj.action,
-                        args))
+                evJob = service.getJob(handler_name, {'event_args': args, 'event_type':event_type})
+                evJob.execute()
```
zaibon: also there is a reason we don't do the execute here. this block completly the daemon during execution.
Please use the `self._workers.apply_async`

lib/JumpScale/baselib/atyourservice81/AtYourServiceDaemon.py
```diff
@@ -212,7 +217,14 @@ def run(self):
 
                     now = j.data.time.epoch
                     for action_name, recurring_obj in service.model.actionsRecurring.items():
-
+                        deptree = service._build_actions_chain(action=action_name)
+                        ### PLEASE EXPLAIN
```
zaibon: not sure what's the goal here, some comment could help indeed

lib/JumpScale/baselib/atyourservice81/AtYourServiceDaemon.py
```diff
@@ -212,7 +217,14 @@ def run(self):
 
                     now = j.data.time.epoch
                     for action_name, recurring_obj in service.model.actionsRecurring.items():
-
+                        deptree = service._build_actions_chain(action=action_name)
```
rkhamis: in a scenario where we have a service action depending on another action that is not defined in its service, this won't be good enough.

imagine:
 - start depends on install
 - service that has start doesn't have install but depends on other services that define install
 - _build_actions_chain will show that start is good to go

docs/Howto/GEODNS/geodns.md
```diff
@@ -0,0 +1,48 @@
...
+
+```python
+domain_manager = j.sal.domainmanager.get(cuisine)
+domain_manager.ensure_domain("gig.com", serial=3, ttl=600)
+```
+## adding **A** record
+
+
+```python
+domain_manager.add_record("gig.com", "www", "a", "123.45.123.1")
```
zaibon: Why don't we use the `domain.add_a_record` method here ?

docs/Howto/GEODNS/geodns.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## adding **cname** record
+```python
+domain_manager.add_record("gig.com", "grid", "cname", "www")
```
zaibon: Why don't we use the `domain.add_cname_record` method here ?

docs/Howto/GEODNS/geodns.md
```diff
@@ -0,0 +1,48 @@
...
+  a_records = domain_manager.get_record("gig.com", "a")
+```
+
+## getting **cname** record
+```python
+  cname_records = domain_manager.get_record("gig.com", "cname")
+```
+## deleting **A** record
+```python
+domain_manager.del_record("gig.com", "a", "www", full=True)
```
zaibon: Why don't we use the `domain.del_a_record` method here ?

docs/Howto/GEODNS/geodns.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+## deleting **cname** record
+```python
+domain_manager.del_record("gig.com", "cname", "grid", full=True)
```
zaibon: Why don't we use the `domain.del_cname_record` method here ?

PR #715, opened 2017-01-12
8.2.0 fixes

rkhamis:


lib/JumpScale/sal/fs/SystemFS.py
```diff
@@ -1353,7 +1353,7 @@ def statPath(self, path):
         if path is None:
             raise TypeError('Path is None in system.fs.statPath')
         try:
-            return os.stat(path)
+            return os.stat(path, follow_symlinks=follow_symlinks)
```
rkhamis: you need to add `follow_sysmlinks=True` in line 1349

PR #735, opened 2017-02-13
8.2.0 fixes

abdulrahmantkhalifa:


lib/JumpScale/clients/ssh/SSHClient.py
```diff
@@ -103,6 +104,7 @@ def client(self):
                 print("could not load pkey")
                 pass
 
+            paramiko.util.log_to_file("/tmp/paramiko.log")
```
zaibon: here we should not use hardcoded `/tmp` but use `$TMPDIR` instead

PR #751, opened 2017-03-02
cuisine portal fixes for mac

abdulrahmantkhalifa:
- branch hardcoded in portal changed to use JSBRANCH varaible instead
- base64 in mac uses -D instead of -d
- jumpscale on macosx doesnt have write access on /etc , changed
  startup.sh to bindir.

### DIFF
--- a/lib/JumpScale/tools/cuisine/CuisineProcessManager.py
+++ b/lib/JumpScale/tools/cuisine/CuisineProcessManager.py
@@ -9,7 +9,7 @@ base = j.tools.cuisine._getBaseClass()
 class ProcessManagerBase(base):
 
     def __init__(self, executor, cuisine):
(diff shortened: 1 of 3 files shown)


lib/JumpScale/tools/cuisine/CuisineProcessManager.py
```diff
@@ -9,7 +9,7 @@
 class ProcessManagerBase(base):
 
     def __init__(self, executor, cuisine):
-        self.startupfile = '/etc/startup.sh'
+        self.startupfile = "%s/startup.sh" % j.dirs.BINDIR
```
ashraffouda: this needs to be also refactored in ayscockpit/actions.py to use the same path for the startup script

PR #815, opened 2017-04-09
Blueprintcontent2string

xmonader:


zaibon:
If we change the way blueprint content should be sent to AYS, make sure to go over all the component that does that and update, AYS CLI, portal macros. etc...

rkhamis:
@zaibon that's all already done in the PR

PR #816, opened 2017-04-10
jobs and runs cleanup

alichaddad:


apps/atyourservice/main.py
```diff
@@ -31,6 +31,21 @@
 @click.option('--port', '-p', default=5000, help='listening port')
 @click.option('--debug', default=False, is_flag=True, help='enable debug logging')
 def main(host, port, debug=False):
+    async def cleanup(sanic, loop):
+        while True:
+            sleep = 60
+            runs = j.core.jobcontroller.db.runs.find()
+            limit = int(j.data.time.getEpochAgo("-2h"))
+            for run in runs:
+                if (run.dbobj.state == 'error' or run.dbobj.state == 'ok'):
```
xmonader: How about state in ['error', 'ok'] and lose the parens?

apps/atyourservice/main.py
```diff
@@ -31,6 +31,21 @@
...
+                    j.core.jobcontroller.db.runs.delete(state=run.dbobj.state, repo=run.dbobj.repo, toEpoch=limit)
+            jobs = j.core.jobcontroller.db.jobs.find()
+            for job in jobs:
+                if (job.state == 'error' or job.state == 'ok'):
```
xmonader: same as above^

apps/atyourservice/main.py
```diff
@@ -31,6 +31,21 @@
 @click.option('--port', '-p', default=5000, help='listening port')
 @click.option('--debug', default=False, is_flag=True, help='enable debug logging')
 def main(host, port, debug=False):
+    async def cleanup(sanic, loop):
+        while True:
+            sleep = 60
+            runs = j.core.jobcontroller.db.runs.find()
+            limit = int(j.data.time.getEpochAgo("-2h"))
```
xmonader: getEpochAgo returns int maybe we can lose the casting here.

PR #834, opened 2017-04-14
change how we generate the node name from its mac address

zaibon:


lib/JumpScale/sal/g8os/atyourservice/StoragePool.py
```diff
@@ -22,10 +22,15 @@ def create(self, aysrepo):
 
     @property
     def _node_name(self):
+        def is_valid_nic(nic):
+            for exclude in ['zt', 'core', 'kvm', 'lo']:
+                if nic['name'].startswith(exclude):
+                    return False
+            return True
+
         for nic in self._obj.node.client.info.nic():
```
grimpy: you are not using the is_valid_nic method, I suggest you you filter method

PR #877, opened 2017-05-04
Improve sal for containers

grimpy:
Signed-off-by: Jo De Boeck <deboeck.jo@gmail.com>


lib/JumpScale/sal/g8os/Container.py
```diff
@@ -1,18 +1,46 @@
...
+        for container in self.node.client.container.list().values():
+            try:
+                containers.append(Container.from_containerinfo(container, self.node))
+            except ValueError:
+                # skip containers withouth tags
+                pass
+        return containers
+
+    def get(self, name):
+        containers = self.node.client.container.find(name)
```
zaibon: we should probably raise in case we found multiple containers with the same name, cause that's not something we supports.

grimpy: Address in updated PR

PR #886, opened 2017-05-09
Remove creation of StorageCluster StoragePool

grimpy:
StorageCluster is a umbrella service which will create the child
services inside the service itself

Removed slaveof/master references from storageserver as this will not be
used by a storagecluster

Signed-off-by: Jo De Boeck <deboeck.jo@gmail.com>


lib/JumpScale/sal/g8os/StorageCluster.py
```diff
@@ -1,36 +1,37 @@
...
+        self.nodes = nodes or []
         self.filesystems = []
         self.storage_servers = []
-        self.disk_type = None
-        self.has_slave = None
+        self.disk_type = disk_type
         self._ays = None
 
     @classmethod
     def from_ays(cls, service):
```
FastGeert: I think this method needs to be moved to the AYS template. Otherwise we are mixing layers.

g8os <== g8sos_pyclient <== js client <== js sal <== ays

PR #892, opened 2017-05-10
added code to remove partition in the delete of storage pool

abdulrahmantkhalifa:
relates to issue https://github.com/g8os/resourcepool/issues/175


lib/JumpScale/sal/g8os/StoragePool.py
```diff
@@ -87,7 +87,9 @@ def delete(self, zero=True):
             self.umount()
         if zero:
             for device in self.devices:
-                self._client.system('dd if=/dev/zero bs=1M count=500 of={}'.format(device))
+                disk = device.split("/")[-1]
+                self._client.disk.rmpart(disk[:-1], int(disk[-1]))
+                self._client.system('dd if=/dev/zero bs=1M count=500 of={}'.format(device[:-1]))
```
grimpy: please use a cleaner way to strip the parition nr from the disk name,
you should will want to wait on this system call by adding a .get() behind it

PR #893, opened 2017-05-10
Fixes #887

### LINKED ISSUE #887: Failed to update actor when adding new parameters
The service doesn't get created when using a blueprint with the updated parameters, producing the error `struct has no such member` for the new parameters.

rkhamis:
+ AYS CLI waits for server to be started instead of crashing (on port 5000)

### DIFF
--- a/lib/JumpScale/baselib/atyourservice81/lib/Service.py
+++ b/lib/JumpScale/baselib/atyourservice81/lib/Service.py
@@ -269,9 +269,14 @@ class Service:
     def _check_args(self, actor, args):
         """ Checks whether if args are the same as in instance model """
         data = j.data.serializer.json.loads(self.model.dataJSON)
-        for key, value in args.items():
-            sanitized_key = j.data.hrd.sanitize_key(key)
-            if sanitized_key in data and data[sanitized_key] != value:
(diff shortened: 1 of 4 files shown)


lib/JumpScale/baselib/atyourservice81/lib/Service.py
```diff
@@ -517,8 +523,10 @@ def processChange(self, actor, changeCategory, args={}, reschedule=False):
         # self.logger.debug('process change for %s (%s)' % (self, changeCategory)
 
         if changeCategory == 'dataschema':
-            # We use the args passed without change
-            pass
+            # update arg values
```
zaibon: I'm not sure we want to force this here.
What we used to do it let this up to the service implementation, cause there are some case where some of the args can't be changed.

rkhamis: I think this should be the default behavior.
We can make it that a service can change that behavior.

zaibon: the problem with this approach is that if you change it here. The service has no way to have the old and new args to see which one he needs to keep. Or do you have a way to allow that ?

PR #895, opened 2017-05-11
change run retry behavior

zaibon:
When a run fails, we retry the full run instead of just the failed
action. Like this we don't need to manually recreate a new run after the
failed action has been solved


lib/JumpScale/baselib/atyourservice81/lib/RunScheduler.py
```diff
@@ -132,55 +131,29 @@ def current_run(self):
...
-        @param service: service object
-        @param action_name: name of the action to retry
-        """
-        async def do_retry(service, action_name):
-
-            service_key = service.model.key
-            action = service.model.actions[action_name]
+    async def _retry(self, run):
+        async def do_retry(run):
+            import ipdb; ipdb.set_trace()
```
grimpy: ipdb ?

lib/JumpScale/baselib/jobcontroller/RunStep.py
```diff
@@ -56,6 +56,10 @@ def _fake_exec(self, job):
         jobs = []
         for job in self.jobs:
 
+            # don't re-execute succefull jobs
```
FastGeert: succesfull (with s)

shellcmds/ays
```diff
@@ -263,7 +263,8 @@ def show(role, name):
 @service.command()
 @click.option('--role', '-r', default="", help='optional role of the service')
 @click.option('--name', '-n', default="", help='optional name of the service')
-def state(role, name):
+@click.option('--state', '-s', default="", help='display all service action that have the specified state')
```
FastGeert: actions (with s)

Issue #842, opened 2017-04-23
AYS: actor templates file watching.


zaibon:
Currently AYS reload the actor templates every minute to make sure it always has the last version of the actor templates loaded. This is not really efficient. It would be better to use something like inotify to receive event when the file actually get change instead of just reloading every minutes.

PR #909, opened 2017-05-17
imporvements for loadings aysrepos actors templaterepos

xmonader:


zaibon:
I don't specially like the fact that we have to use `find`.  
 Also I would like to see how https://github.com/Jumpscale/jumpscale_core8/issues/842 will impact on the overall perfs

rkhamis:
@zaibon already the `find` cut down the scanning time to less than a third.

I'll merge for optimizati

lib/JumpScale/baselib/atyourservice81/lib/TemplateRepo.py
```diff
@@ -1,65 +1,22 @@
 from JumpScale import j
 from JumpScale.baselib.atyourservice81.lib.ActorTemplate import ActorTemplate
 import asyncio
+import os
```
rkhamis: Not sure we need this import

zaibon:
but isn't this commit will conflict with #842 ?
From what I see #842 will remove this code altogether

lib/JumpScale/baselib/atyourservice81/lib/AtYourServiceRepo.py
```diff
@@ -70,22 +70,14 @@ def _searchAysRepos(self, path):
...
-            if base in [".git", ".hg", ".github"]:
-                return False
-            depth = len(j.sal.fs.pathRemoveDirPart(path, arg[0]).split("/"))
-            # print("%s:%s" % (depth, j.sal.fs.pathRemoveDirPart(path, arg[0])))
-            if depth < 6:
-                return True
-            return False
+        cmd = """find %s \( -wholename '*.ays' \) -exec readlink -f {} \;""" % (path)
+        rc, out, err = j.sal.process.execute(cmd, die=False, showout=False)
+        for reponame in out.splitlines():
```
rkhamis: Maybe we should add a strip and a check that the line is not empty here as well

PR #910, opened 2017-05-17
watch the ays repo using inotify

mie00:


zaibon:
we also need to handle the templateRepo : https://github.com/Jumpscale/jumpscale_core8/blob/8.2.0/lib/JumpScale/baselib/atyourservice81/lib/TemplateRepo.py#L6
and the templates themself:
https://github.com/Jumpscale/jumpscale_core8/blob/8.2.0/lib/JumpScale/baselib/atyourservice81/lib/TemplateRepo.py#L202

lib/JumpScale/baselib/atyourservice81/lib/TemplateRepo.py
```diff
@@ -1,7 +1,24 @@
...
-import os
+import inotify
+
+
+def get_root_template_repo_if_relevant(path):
+    template_repo = None
+    if 'templates' in path:
+        template_repo = path.split("templates")[0]
+    elif 'tests' in path:
+        template_repo = path.split("tests")[0]
```
zaibon: Why do we match on `tests` ?

rkhamis: Otherwise no testsuites would run

rkhamis:
pyinotify is forked here: https://github.com/gigforks/PyInotify

PR #918, opened 2017-05-22
820nologgers

xmonader:
Trying to have 1 general logger for jumpscale.

### DIFF
--- a/lib/JumpScale/baselib/jobcontroller/Job.py
+++ b/lib/JumpScale/baselib/jobcontroller/Job.py
@@ -60,7 +60,6 @@ def _execute_cb(job, future):
             service_action_obj.state = 'ok'
         if job.service:
             job.service.model.dbobj.state = 'ok'
-
         job.logger.info("job {} done sucessfuly".format(str(job)))
     if service_action_obj.period > 0:  # recurring action.
         job.model.delete()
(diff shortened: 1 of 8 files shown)


lib/JumpScale/baselib/jobcontroller/Job.py
```diff
@@ -116,7 +116,7 @@ def __init__(self, model):
         self._future = None
         self.saveService = True
         self._sourceLoader = None
-        self.logger = j.logger.get('j.jobcontroller.job.{}'.format(self.model.key))
+        self.logger = j.logger.get('j.jobcontroller.job.{}'.format(self.model.key), new=True)
```
zaibon: Why do we still create a special logger for jobs ? The idea was to use the same logger as for the rest so we have the logs of all the modules

lib/JumpScale/core/logging/LoggerFactory.py
```diff
@@ -136,36 +136,16 @@ def init(self, mode, level, filter=[]):
         if filter:
             self.handlers.consoleHandler.addFilter(ModuleFilter(filter))
 
-    def get(self, name=None, enable_only_me=False):
+    def get(self, name=None, new=False):
```
zaibon: I don't like this `new` argument. It's obscure and not documented.
Why don't just have name to None by default and if name is None we return the root logger
Also the root logger should be now called `jumpscale` not just `j` anymore.

lib/JumpScale/sal/fs/SystemFS.py
```diff
@@ -23,6 +24,9 @@
 # import JumpScale.baselib.codetools #requirement for parsePath
 # from text import Text
 
+logger = logging.getLogger("j.sal.fs")
+logger.addHandler(logging.NullHandler())
```
zaibon: Don't get that either ? why this lib should have a different logger ?

